PR #2, opened 2021-07-12
Package and CI config changes

ageorgou:
Some changes and enhancements:

- Minor change to `setup.cfg` which I found necessary for installing the package successfully
- Ignore some more files
- Set up basic tests on GitHub Actions (if we decide to go with that)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -3,6 +3,7 @@ name: Run tests
 on:
   push:
   pull_request:
+    branches: [main, develop]
 
 jobs:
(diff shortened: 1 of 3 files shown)


ageorgou:
The CI is failing because there are no tests added on this branch yet :)

pweschmidt (commented):
Following the discussion in the mgt meeting today: we should set up GitHub actions rather than Travis build/test. I can dig up what we did at QNice

pweschmidt:
My bad - I didn't see that there is a GitHub action already

Issue #1, opened 2021-07-08
Implement evolution of a fixed size population over a defined period of time


pweschmidt:
I would propose this to be our initial use case:
From a determined start date until a determined end date (e.g. 1989 - 2071), define a total population size (e.g. 100000) and

- Set age randomly at the start (not yet borns will get a negative age)
- Set a fixed age limit (e.g. 70 years)
- Run the algorithm from start to end date
- At the end, how many are still alive at end date and what is the age distribution?

It could help us set up the architecture for the basic modelling end-to-end workflow:

- how to configure a model run: parameters such as start/end dates, time increments, population sizes as well as "logistics" parameters such as where the output is being stored (and potentially HPC parameters?)
- how to run the model itself
- how to produce intermediate/log output
- how to produce final output

PR #6, opened 2021-07-22
Compute number of people alive across simulation

ageorgou:
Towards #1.

### DIFF
--- a/src/tests/test_simulation.py
+++ b/src/tests/test_simulation.py
@@ -3,6 +3,7 @@ from datetime import timedelta, date
 import pytest
 
 from hivpy import run_simulation, SimulationConfig, SimulationException
+from hivpy import population
 from hivpy.population import Population
 
 @pytest.fixture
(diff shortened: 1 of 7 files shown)


src/tests/test_simulation.py
```diff
@@ -58,3 +59,18 @@ def test_error_tracking_nonexistent(sample_pop):
...
+
+def test_death_occurs():
+    """Check that the number of people alive always decreases."""
+    # FIXME This will not necessarily be true once we add in births
+    population = Population(10000, date.today())
+    start = population.date
+    step = timedelta(days=30)
+    end = start + 100 * step
+    config = SimulationConfig(start, end, step, ['num_alive'])
+    pop, results = run_simulation(population, config)
```
pweschmidt: that and the line above would need to change once you merged the code from my branch. E.g. run_simulation takes only one config (the simulation config) and Population is a class member of SimulationHandler

pweschmidt (commented):
The simulation looks ok. I left a comment for the test case - as this code would need to change if the changes to the Configs are accepted.

pweschmidt:
The test needs to be modified slightly with the latest changes:
instead of 
.
.
.
config = SimulationConfig(start, end, step, ['num_alive'])
pop, results = run_simulation(population, config)

it is now
config = SimulationConfig(population_size, start, end, ['num_alive']
pop, results = run_simulation(config)

PR #7, opened 2021-07-22
Command interface

pweschmidt:
Hey Anastasis
here is the stuff we just talked about. I removed core.py as discussed


src/hivpy/cli.py
```diff
@@ -0,0 +1,38 @@
...
+    parameter_filepath = args.parameters
+
+'''
+Running a HIV model simulation
+Assuming there is a hivpy.conf file the command is
+$run_model hivpy.conf
+'''
+def run_model():
+    parser = argparse.ArgumentParser(description="run a simulation")
+    parser.add_argument("input", type=pathlib.Path, help="run_model -i config.conf")
```
ageorgou: The help string has an extra `-i`, but also I think it would be better to describe the particular argument (users will get the suggested usage string automatically if you use the wrong syntax).

src/hivpy/experiment.py
```diff
@@ -1,34 +1,59 @@
...
+        logging.basicConfig(filename=self.logfile, level=LEVELS[self.loglevel])
+        logging.info("starting experiment")
+
+
+
+def create_experiment(experiment_param):
+    start_date = date(1980,1,1)
+    end_date = date(2040, 12, 31)
+    interval = timedelta(days=30)
+    population_size = 1000
```
ageorgou: If we're overwriting these anyway, we can remove this?

src/hivpy/experiment.py
```diff
@@ -1,34 +1,59 @@
...
+    try:
+        start_date = date(int(experiment_param['START_YEAR']),1 ,1 )
+        end_date = date(int(experiment_param['END_YEAR']),12, 31)
+        population_size = int(experiment_param['POPULATION'])
+        interval = timedelta(days = int(experiment_param['TIME_INTERVAL_DAYS']))
+    except RuntimeError as err:
```
ageorgou: I'm not sure this will catch what we want. For example, `int` and `date` both raise `ValueError` if their arguments are not appropriate, so it may be better to catch that instead of `RuntimeError`.

src/hivpy/experiment.py
```diff
@@ -1,34 +1,59 @@
...
+    'WARNING': logging.WARNING,
+    'ERROR': logging.ERROR,
+    'CRITICAL': logging.CRITICAL
+}
 
 @dataclass
-class ExperimentConfig:
-    """Configuration parameters covering all stages of an experiment."""
-    population_size: int
-    simulation: SimulationConfig
```
ageorgou, line 11: My initial intention for `ExperimentConfig` was that it would hold together all the configs we read from one file (for the simulation, the output, any post-processing etc),  that's why it was kept distinct from `SimulationConfig`. If you think that's not useful though, happy to remove it.

src/hivpy/experiment.py
```diff
@@ -1,34 +1,59 @@
...
-        return Population(self.population_size, self.simulation.start_date)
+    # Dummy values for now
+    return SimulationConfig(population_size, start_date, end_date, interval)
 
+def create_output(output_param):
+    outputdir = output_param['OUTPUT_DIRECTORY']
+    logfilename = output_param['LOGOUTPUT_NAME']
+    log_level = output_param['LOG_LEVEL']
+    logpath = os.path.join(outputdir, logfilename)
+    return OutputConfig(outputdir, logpath, log_level)
```
ageorgou, line 27: This can definitely wait until later, but we should discuss if we want to have some default values for these, so the config file doesn't need to provide them all (or perhaps it would be better to have everything explicitly)

ageorgou (changes requested):
Looks good, main question is about whether to keep a class that brings together all the different configs (e.g. in case we need information from multiple of them, or to pass fewer arguments around for convenience)

src/hivpy/cli.py
pweschmidt: Removed the -i. I will see how the help in the command line can provide more info on expected arguments

src/hivpy/experiment.py
pweschmidt: I remember why I put them there: it is to ensure some default values in case none have been provided in the config file. But, yes, there should be a different way of doing this

src/hivpy/experiment.py
pweschmidt: will replace it with ValueError

src/hivpy/experiment.py
pweschmidt: Maybe a single Config class comprising both simulation and output config would be good - which would also be a place to have sensible default values in case none are provided in a config file.

src/hivpy/experiment.py
ageorgou: Right, I see. If we want that, then the exception block should also catch `KeyError` (to deal with values not specified in the config file)

src/hivpy/cli.py
ageorgou: I think it's done by default if the user doesn't give the right arguments: the expected usage is automatically assembled and displayed. I don't think you need to do anything extra.

The `help` string is meant to describe each input argument (like "a file with the parameters of the simulation") so it can be displayed in the overall help/usage message.

ageorgou (approved):
The tests need some updates from these but that can be changed at the next stage.

PR #12, opened 2021-11-10
Enhancement/add age gender

mmcleod89:
Adds demographic class for population characteristics. 

Add class for sex, assigned to the population at initial time. 

Add classes for continuous and stepwise age distributions. 
Continuous age distributions at the moment use a linear-exponential model which fits the three example scenarios quite well with less pathological behaviour than polynomials. 

Adds scipy as a dependency.

### DIFF
--- a/src/hivpy/demographics.py
+++ b/src/hivpy/demographics.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
 from pandas.api.types import CategoricalDtype
 from scipy.interpolate import interp1d
@@ -39,10 +41,18 @@ class StepwiseAgeDistribution:
     # Generate cumulative probabilites from stepwise distribution
     def _genCumulativeProb(self):
         N = self.probability_list.size
+        print("IS THIS EVEN WORKING")
+        if(not np.isclose(sum(self.probability_list), 1.0, atol=1e-10, equal_nan=False)):
+            warnings.warn("Age probability distribution does not sum to one.")
         CP = np.array([0.0]*N)
         CP[0] = self.probability_list[0]
         for i in range(1, N):
             CP[i] = CP[i-1] + self.probability_list[i]
+            if(CP[i] > 1.0):
+                warnings.warn("Cumulative probability grown too large; capping at 1.0.")
+                CP[i] = 1.0
+        # make sure CP is exactly 1.0 in case of numerical error
+        CP[N-1] = 1.0
         return CP
 
     # Generate Age With Stepwise Distribution
@@ -67,11 +77,16 @@ class StepwiseAgeDistribution:
 
 
 class ContinuousAgeDistribution:
-    # example distribution: linexp
-    # y = (mx + x)*exp(A(x-B))
-    # This fits all three example cases quite neatly
-    # We only need the cumulative distribution i.e. the integral of the function
-    def integratedLinexp(x, m, c, A, B):
+    """
+    Class to handle age distributions using a continuous probability density.
+    Linear-Exponential function is used for the example distribution.
+
+    y = (mx + c)*exp(A(x-B))
+
+    This fits all three example cases quite neatly.
+    """
+
+    def _integratedLinexp(self, x, m, c, A, B):
         return np.exp(A*(x-B))*(m*x+c - m/A)/A
 
     # Example parameters
@@ -79,19 +94,25 @@ class ContinuousAgeDistribution:
     modelParams2 = [-1.03e-3, 7.45e-2, -1.12e-3, 8.47]
     modelParams3 = [-1.15e-3, 8.47e-2, 2.24e-3, 2.49e1]
 
-    def __init__(self, min_age, max_age, cpd):
+    def __init__(self, min_age, max_age, modelParams):
         self.min_age = min_age
-        self.max_age = max_age
-        self.cpd = cpd
+        model_age_limit = -modelParams[1]/modelParams[0]
+        if(max_age > model_age_limit):
+            warnings.warn("Max age exceeds the maximum age limit for this model (negative probability).\
+                            Adjusting max age to " + model_age_limit)
+            self.max_age = model_age_limit
+        else:
+            self.max_age = max_age
+        self.cpd = lambda x: self._integratedLinexp(x, *modelParams)
 
     @classmethod
     def selectModel(cls, inc_cat):
         if(inc_cat == 1):
-            return cls(-68, 65, lambda x: cls.integratedLinexp(x, *cls.modelParams1))
+            return cls(-68, 65, cls.modelParams1)
         elif(inc_cat == 2):
-            return cls(-68, 65, lambda x: cls.integratedLinexp(x, *cls.modelParams2))
+            return cls(-68, 65, cls.modelParams2)
         else:
-            return cls(-68, 65, lambda x: cls.integratedLinexp(x, *cls.modelParams3))
+            return cls(-68, 65, *cls.modelParams3)
 
     # Generate ages using a (non-normalised) continuous cumulative probability distribution
     # Given an analytic PD, this should also be analytically defined
@@ -107,7 +128,7 @@ class ContinuousAgeDistribution:
         NormX = np.linspace(self.min_age, self.max_age, 11)
         NormY = NormDist(NormX)
         NormY[0] = 0.0
-        NormY[10] = 1.0
+        NormY[10] = 1.0  # fix the start and end values in case of numerical errors
         NormInv = interp1d(NormY, NormX, kind='cubic')
 
         # generate N random numbers in (0,1) and convert to ages
--- a/src/hivpy/demographics.py
+++ b/src/hivpy/demographics.py
@@ -120,7 +120,9 @@ class DemographicsModule:
 
     def __init__(self, **kwargs):
         params = {
-            "female_ratio": FEMALE_RATIO
+            "female_ratio": FEMALE_RATIO,
+            "use_stepwise_ages": USE_STEPWISE_AGES,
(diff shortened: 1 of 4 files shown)


src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+    stepwise_boundaries = np.array([-68, -55, -45, -35, -25, -15, -5, 5, 15,
+                                    25, 35, 45, 55, 65])
+
+    def __init__(self, ages, probabilities):
+        self.probability_list = probabilities
+        self.age_boundaries = ages
+
+    @classmethod
+    def selectModel(cls, inc_cat):
+        if(inc_cat == 1):
```
ageorgou: Maybe add an assertion that `inc_cat in [1, 2, 3]`, to avoid accidentally using the third one?

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+            return cls(cls.stepwise_boundaries, cls.stepwise_model1)
+        elif(inc_cat == 2):
+            return cls(cls.stepwise_boundaries, cls.stepwise_model1)
+        else:
+            return cls(cls.stepwise_boundaries, cls.stepwise_model1)
```
ageorgou: Wait, shouldn't these use the other distributions?

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+
+    # Generate cumulative probabilites from stepwise distribution
+    def _genCumulativeProb(self):
+        N = self.probability_list.size
+        print("IS THIS EVEN WORKING")
```
ageorgou: Well, is it? 😀

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+        if(not np.isclose(sum(self.probability_list), 1.0, atol=1e-10, equal_nan=False)):
+            warnings.warn("Age probability distribution does not sum to one.")
+        CP = np.array([0.0]*N)
```
ageorgou: Numpy also has a method for this: `np.zeros(N)`

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+        CP[0] = self.probability_list[0]
+        for i in range(1, N):
+            CP[i] = CP[i-1] + self.probability_list[i]
+            if(CP[i] > 1.0):
+                warnings.warn("Cumulative probability grown too large; capping at 1.0.")
+                CP[i] = 1.0
+        # make sure CP is exactly 1.0 in case of numerical error
+        CP[N-1] = 1.0
```
ageorgou: We can use built-in methods for this computation:
```suggestion
        CP = self.probability_list.cumsum()
        CP[CP > 1.0] = 1.0
        CP[N-1] = 1.0
```
Although I would lean towards being more restrictive, and raising an error at this point (or at construction) if the cumulative distribution is not correct. In that case, we wouldn't need the corrections either.

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+    stepwise_model1 = np.array([0.18, 0.165, 0.144, 0.114, 0.09, 0.08, 0.068,
+                                0.047, 0.036, 0.027, 0.021, 0.016, 0.012])
+    stepwise_model2 = np.array([0.15, 0.13, 0.12, 0.11, 0.1, 0.09, 0.08, 0.065,
+                                0.048, 0.04, 0.03, 0.021, 0.016])
+    stepwise_model3 = np.array([0.128, 0.119, 0.113, 0.104, 0.097, 0.09, 0.081,
+                                0.074, 0.06, 0.05, 0.038, 0.026, 0.02])
+    stepwise_boundaries = np.array([-68, -55, -45, -35, -25, -15, -5, 5, 15,
+                                    25, 35, 45, 55, 65])
+
+    def __init__(self, ages, probabilities):
```
ageorgou, line 30: Could add a check here that
```suggestion
    def __init__(self, ages, probabilities):
        assert len(ages) == len(probabilities) + 1
```

to ensure the generation works correctly later.

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+        cpd = self._genCumulativeProb()
+        p0 = 0.0
+        rands = np.random.rand(N)
+        ages_out = np.array([0.0]*N)
+        for i in range(0, cpd.size):
+            # in order to vectorise this method we create a mask of values
+            # that need to change for each age group
+            prob_mask = (p0 < rands) & (rands <= cpd[i])
+            ages_out += ((self.age_boundaries[i]
+                          + (rands - p0)/(cpd[i]-p0)
```
ageorgou, line 75: So we're sampling from within the bin using the existing `rands`... I guess that's the same as sampling uniformly inside the age range, right? Since `rands` is uniform overall and we're conditioning on being inside a particular bin.

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+    Class to handle age distributions using a continuous probability density.
+    Linear-Exponential function is used for the example distribution.
+
+    y = (mx + c)*exp(A(x-B))
+
+    This fits all three example cases quite neatly.
+    """
+
+    def _integratedLinexp(self, x, m, c, A, B):
+        return np.exp(A*(x-B))*(m*x+c - m/A)/A
```
ageorgou, line 103: This doesn't match the equation in the docstring exactly - presumably this is the intended one?

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+                warnings.warn("Cumulative probability grown too large; capping at 1.0.")
+                CP[i] = 1.0
+        # make sure CP is exactly 1.0 in case of numerical error
+        CP[N-1] = 1.0
+        return CP
+
+    # Generate Age With Stepwise Distribution
+    # the size of ages should be one larger than the cumulative probabilty
+    # so that each bucket has an upper and lower bound.
+    def genAges(self, N):
```
ageorgou, line 60: Making this a docstring:
```suggestion
    def genAges(self, N):
    """Generate Age With Stepwise Distribution.
    
    The size of ages should be one larger than the cumulative probabilty
    So that each bucket has an upper and lower bound.
    """
```

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+
+
+class ContinuousAgeDistribution:
+    """
+    Class to handle age distributions using a continuous probability density.
+    Linear-Exponential function is used for the example distribution.
+
+    y = (mx + c)*exp(A(x-B))
+
+    This fits all three example cases quite neatly.
```
ageorgou, line 90: Can add a brief explanation of the meaning of the parameters?

src/hivpy/demographics.py
```diff
@@ -1,19 +1,149 @@
...
+    def genAges(self, N):
+        # Normalise distribution over given range
+        C = self.cpd(self.min_age)
+        M = 1/(self.cpd(self.max_age)-self.cpd(self.min_age))
+
+        def NormDist(x):
+            return M*(self.cpd(x) - C)
+
+        # sample and invert the normalised distribution (in case analytic inverse in impractical)
+        NormX = np.linspace(self.min_age, self.max_age, 11)
```
ageorgou: I would use e.g. `num_points = 11` to avoid the magic number here and in the upper limit of `NormY`.

ageorgou (commented):
This looks good, and I like the tests. A few comments but nothing major.

If we want to be strict about adherence to style, we should rename the functions, e.g. `_gen_cumulative_prob` instead of `_genCumulativeProb`. But we can discuss this!

We should also have `Population` call `initialize_ages` from the module!

src/hivpy/demographics.py
mmcleod89: I forgot that was in there! Well, yes and no... The function works, but the point of that line was actually to test whether I could see printed output from running pytests, which I can't. I don't know if pytests redirects it somewhere where it can be seen, but I couldn't find it.

src/hivpy/demographics.py
mmcleod89: Yes, within each age bracket the sampling is uniform; this is to match the SAS code which uses a uniform distribution within each bin (rather than e.g. linear interpolation or something).

src/hivpy/demographics.py
mmcleod89: The docstring describes the probability density function, because that is the most intuitive to work with and select a model for. But we don't actually need the probability density to do the calculation, we need the cumulative distribution i.e. the integral of the probability density. This function is the integral (hence integratedLinexp) rather than the probability density itself. The probability density function doesn't actually appear in this class explicitly because we never call it.

src/hivpy/demographics.py
ageorgou: Yes, pytest captures (and hides) the output by default. Running with `pytest -s`  (or `pytest --capture=no`) should show the output. You can probably configure VS Code to do that too, if you're running them from within the IDE, but not sure how.

src/hivpy/demographics.py
ageorgou: Yep, that makes sense. I was just confused because I expected to use another uniform draw for the sampling inside the bin, but there's no reason not to reuse the existing one.

src/hivpy/demographics.py
ageorgou: Right, that makes sense, I'd missed that.

ageorgou:
Thanks for the updates, looks good to me now. Do you want to add the call to the age initialisation in `Population` for completeness, or leave that for another PR?

ageorgou (approved):

PR #15, opened 2021-11-22
Refactor config creation from file

### LINKED ISSUE #14: Refactor extracting data from Config file
At the moment we use almost 2 constructors for classes like SimulationConfig. This can be optimised/refactored so that the constructors or a method of the classes parse/take the data from the config parameters

ageorgou:
Fixes #14.

Moving some things around. This should also make it easier if we want to change the format of the config file.

I wonder if it would be better for each config to live in the corresponding module, e.g. `SimulationConfig` in `simulation.py` along with the simulation handler, etc.

### DIFF
--- a/src/hivpy/cli.py
+++ b/src/hivpy/cli.py
@@ -2,7 +2,8 @@ import argparse
 import configparser
 import pathlib
 
-from .experiment import create_experiment, run_experiment
+from .config import ExperimentConfig
+from .experiment import run_experiment
 
 '''
 Maybe we may want to register parameters/properties in future?
@@ -28,7 +29,7 @@ def run_model():
     config = configparser.ConfigParser()
     try:
         config.read(conf_filename)
-        experiment_config = create_experiment(config)
+        experiment_config = ExperimentConfig.from_file(config)
         run_experiment(experiment_config)
 
     except configparser.Error as err:
--- a/src/hivpy/cli.py
+++ b/src/hivpy/cli.py
@@ -1,8 +1,8 @@
 import argparse
-import configparser
 import pathlib
 
 from .config import ExperimentConfig
+from .exceptions import ConfigException
 from .experiment import run_experiment
 
 '''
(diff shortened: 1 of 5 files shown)


mmcleod89 (commented):
I don't really understand this refactoring. My problem with the initial structure was that we essentially had a structure where the methods to instantiate each class were held outside of the classes in one file (experiment.py), so we were essentially loading the config into a data-structure, then using these floating methods to copy them across, and then passing them to a data-class as arguments. I think that it would be better to cut out the middle man, and pass the relevant config straight to the appropriate class, so that a) the code is more modular (we don't have all these effective initialisers in one file separate from their classes) and b) we don't need to copy these values and pass them around more than is necessary. 

I think that this refactor is essentially doing the same thing that I was concerned about before. Instead of having functions we have a config class for each class, but otherwise the pattern seems to be the same, with a config for each object existing outside of that object's own class, all in one file (now config.py), and just copying things from an existing config data structure and then passing it onto the initialiser of the object. Why not just pass the relevant config (e.g. simulation_section) to the class directly (in the constructor or some other method)? Although the config classes do some checking of validity, it still seems to be conceptually and oragnisationally clearer to keep these methods together as part of the object's construction rather than having two classes for every kind of object (e.g. Simulation and SimulationConfig)

One area where we do want a config class of its own however is logging (in this branch called output) because that is used to configure a class that is outside of our definition.

ageorgou:
I see benefit in having a separate config object to group the various options needed for e.g. a simulation, even if it doesn't have any dynamic behaviour. If not, you would either need to pass them all individually to the `SimulationHandler`, or grouped together in, say, a dictionary, and then validate that (for missing keys or invalid combinations, etc). It's more of a record with some built-in validation.

I understand the concern about the two-step structure but I think that's a side-effect of reading the options from a file. Having a separate config class also makes it simpler (in my view) to create it in different ways, e.g. by directly passing the values without a text file. Isn't that more modular?

mmcleod89:
But aren't they already grouped together in a dictionary like object? You don't pass the entire config file to e.g. SimulationConfig, just simulation_section, which has dictionary like lookup methods (e.g. simulation_section['START_YEAR']), so all this grouping already exists; the data isn't going directly from the file on disk into the config classes right? So this isn't really a two step structure, it's a three step: file -> config_parser -> config classes -> object class. 

Being able to create the objects in different ways can be useful, but it also seems like that can be achieved without an intermediary class using e.g. class methods to create an object. It seems to me that since the something like SimulationConfig only does work related to the Simulation class and doesn't have utility outside of it then it should really just be part of that class. 

It would also make sense to me to use separate config classes if there were some config superclass that they inherited from that gave them some reusable methods, but as it stands each one needs to be completely rewritten anyway with unique methods so doesn't it make just as much sense for these to be simply methods of the class that they configure rather than separate classes?

ageorgou:
This is now outdated and it is unclear what the best way foward is. Closing, we should reexamine the best solution for the current state of the code as part of addressing the underlying issue (#21).

PR #16, opened 2021-11-30
Enhancement/logging output

mmcleod89:


src/hivpy/demographics.py
```diff
@@ -113,23 +113,24 @@ def __init__(self, min_age, max_age, modelParams):
         self.min_age = min_age
         model_age_limit = -modelParams[1]/modelParams[0]
         if(max_age > model_age_limit):
-            warnings.warn("Max age exceeds the maximum age limit for this model (negative probability).\
-                            Adjusting max age to " + model_age_limit)
+            logging.getLogger("Demographics").warning(f"Max age exceeds the maximum age limit for\
+                                                       this model (negative probability).\
+                                                       Adjusting max age to {model_age_limit}")
```
ageorgou: Very minor - I think there's a space missing here between the lines?

If you want to avoid the continuation `\`, you can also break up string literals over multiple lines and they will be joined automatically:
```suggestion
            logging.getLogger("Demographics").warning(f"Max age exceeds the maximum age limit for "
                                                       f"this model (negative probability). "
                                                       f"Adjusting max age to {model_age_limit}")
```
(I may well have messed up the indentation here)

src/hivpy/experiment.py
```diff
@@ -1,10 +1,58 @@
...
+    def __init__(self, output_params):
+        self.output_dir = os.path.normpath(output_params['OUTPUT_DIR'])
+        self.files = {"Raw": "raw_data.out", "Stats": "summary_statistics.out"}
+
+    def set_file_name(self, fileId, filename):
+        filepath = os.path.join(self.output_dir, filename)
+        self.files[fileId] = filepath
+
+    def _commit_data(self, fileId, data, write_mode):
+        if(not (fileId in self.files)):
```
ageorgou: Personal preference:
```suggestion
        if fileId not in self.files:
```

src/hivpy/experiment.py
```diff
@@ -1,10 +1,58 @@
...
+            raise SimulationException(f"File matching {fileId} does not exist.")
```
ageorgou: I would use a distinct exception type, or rename this to e.g. `HIVException`. Depends on whether we want to distinguish between errors at different stages/of different causes (my general instinct is yes).

src/hivpy/experiment.py
```diff
@@ -1,10 +1,58 @@
...
+        if(isinstance(data, pd.DataFrame)):
+            data.to_csv(self.files[fileId], mode=write_mode)
+        elif(isinstance(data, np.ndarray)):
+            outfile = open(self.files[fileId], write_mode)
+            np.savetxt(outfile, data)
+            outfile.close()
```
ageorgou: Can use a context manager to handle closing of the file at the end or if any errors occur:
```suggestion
            with open(self.files[fileId], write_mode) as outfile:
                np.savetxt(outfile, data)
```
Although here there's not much to protect against.

ageorgou: Do we want this to also be in CSV? If so, we should pass `delimiter=","` to `np.savetxt` (I think it uses spaces by default)

src/hivpy/experiment.py
```diff
@@ -1,10 +1,58 @@
...
+        self._commit_data(self, fileId, data, 'a')
+
+    def overwrite(self, fileId, data):
+        self._commit_data(self, fileId, data, 'w')
+
+    def _gen_statistic(self, f_stat, field):
+        return f_stat(self.simulation_data[field])
+
+    def gen_summary_stats(self):
+        [self._gen_statistic(f_stat, field) for (f_stat, field) in self.summary_stats]
```
ageorgou:
```suggestion
        return [self._gen_statistic(f_stat, field) for (f_stat, field) in self.summary_stats]
```

src/hivpy/config.py
```diff
@@ -15,14 +16,31 @@
...
-        logging.basicConfig(filename=self.logfile, level=LEVELS[self.loglevel])
-        logging.info("starting experiment")
+        # file logging
+        file = path.join(self.log_dir, self.logfile)
+        logging.root.setLevel(logging.DEBUG)
+        file_logger = logging.FileHandler(file, 'w')
+        file_formatter = logging.Formatter('%(asctime)s %(name)-15s %(levelname)-10s %(message)s',
+                                           datefmt='%y-%d-%m %H:%M:%S')
+        file_logger.setFormatter(file_formatter)
+        file_logger.setLevel(logging.DEBUG)
```
ageorgou: Should this be using `fileLogLevel`? (same for the console)

src/tests/test_logging.py
```diff
@@ -0,0 +1,36 @@
...
+    logger.critical(TEST_CRITICAL)
+    console_out = capsys.readouterr().err
+    file_out = open(path.normpath(path.join(d, "hivpy.log")), 'r')
+    file_text = file_out.read()
+    file_out.close()
+    assert TEST_DEBUG in file_text
+    assert TEST_INFO in file_text
+    assert TEST_WARNING in console_out
+    assert TEST_ERROR in console_out
+    assert TEST_CRITICAL in console_out
```
ageorgou:
```suggestion
    assert TEST_WARNING in file_text
    assert TEST_ERROR in file_text
    assert TEST_CRITICAL in file_text
```

ageorgou (commented):
The logging changes all look good, only have some minor suggestions and typos.

I'm not sure about including the statistics just yet while they're not quite functional (unless I'm missing something!). For example, we're missing the initialisation, and the type of `summary_stats` looks more like it's a list/sequence of tuples rather than an `ndarray`. Can keep them for another PR along with a usage example? Alternatively, can complete a quick implementation now and refine it later.

src/hivpy/experiment.py
mmcleod89: I agree; I wasn't sure if we wanted to start making changes to the exceptions file as part of this as we hadn't discussed any error handling yet, but I will add another exception class OutputException for this case as I think we probably will want to be able to distinguish between different cases, although perhaps we will want an HIVException superclass

src/hivpy/experiment.py
mmcleod89: Agreed; I should really have put in a try statement in case the file write went wrong somehow, but a context manager is neater. (I find them a bit confusing because I don't know what classes have __enter__ and __exit__ functionalities to allow them to be used with it.) Also yes it is a good idea to keep the files consistent so I'll make it comma delimited!

ageorgou (approved):

Issue #17, opened 2021-12-10
Death rates in model


pweschmidt:
Implement in the model the mortality of the population in the absence of HIV in the first instance.

Optional: add death rates incl HIV death risk

ageorgou:
Basic implementation done in #27. Follow-up work will be tracked other issues such as #44.

PR #21, opened 2022-01-17
Enhancement/sexual behaviour

mmcleod89:
Very much WIP but useful to make things visible. I have also taken out the probability.py file and replaced with the scipy class, because that most closely resembled the approach I had taken so far with my class, but that may be swapped out for numpy if it turns out to be a better approach when I try to integrate this all into a simulation 
Next steps for me:
- Need to move sexual behaviour stats into a config file and add a randomised selection function 
- Add calculation of num partners to the simulation so we can see it change over time


src/hivpy/population.py
```diff
@@ -36,21 +39,34 @@ def _sample_parameters(self):
...
 
+    def _prob_HIV_initial(self, age):
+        """Completely arbitrary placeholder function for initial HIV presence in population"""
+        return np.clip(self.hiv_a*age**2 + self.hiv_b*age + self.hiv_c, 0.0, 1.0)
+
+    def _initial_HIV_status(self):
+        """Initialise HIV status based on age (& sex?)"""
+        """Assume zero prevalence for age < 15"""
+        hiv_status = self._prob_HIV_initial(self.data["age"]) > np.random.rand(self.size)
+        return hiv_status
```
ageorgou, line 52: In an ideal world (that we don't live in!), all such initialisation functions (and the associated params) would be in an appropriate specific "module" and not in the Population class. I don't know if this would fit better in demographics or a separate hiv module - depends on whether we include more details of transmission dynamics, perhaps.

src/hivpy/population.py
```diff
@@ -81,6 +97,9 @@ def get(self, attribute_name):
         """Get the value of the named attribute at the current date."""
         return self.attributes[attribute_name](self.data)
 
+    def _update_HIV_status(self):
+        """Update HIV status for new transmissions in the last time period"""
+
```
ageorgou, line 103: Similarly, this could go in the sexual behaviour module, and called from `evolve`? Or perhaps it will end up calling a series of update/util functions from other modules, in which case it's useful as a group here.

src/hivpy/sex_behaviour_data.py
```diff
@@ -0,0 +1,179 @@
...
+    [[0.98, 0.02],
+     [0.85, 0.15]],
+
+    [[0.98, 0.02],
+     [0.75, 0.25]]
+])
+
+# Sexual age mixing matrix
+# groups 15-24, 35-34, 45-54, 55-65
+# presumably no partners over 65?
```
ageorgou, line 154: Yes, I believe most parts of the code ignore sexual activity after 65 (but some other metrics are tracked up to 80, or for any age).

src/hivpy/sex_behaviour_data.py
```diff
@@ -0,0 +1,179 @@
...
+                          [0.2,   0.03],
+                          [0.15,  0.02]])
+
+# Sexual behaviour transition matrices
+
+# Male
+# Probability of transitions between male sexual behaviour groups
+# 0 = Zero, 1 = Low, 2 = Medium, 3 = High
+sex_behaviour_trans_male_options = np.array([
+    [[0.995, 0.005, 0.005, 0.00005],
```
ageorgou, line 26: Huh, I thought these had to sum to 1 but it seems they are normalised afterwards in the original model (I think for ease of expression and modification)

src/hivpy/sex_behaviour_data.py
```diff
@@ -0,0 +1,179 @@
...
+
+    [[0.98, 0.02],
+     [0.85, 0.15]],
+
+    [[0.98, 0.02],
+     [0.75, 0.25]]
+])
+
+# Sexual age mixing matrix
+# groups 15-24, 35-34, 45-54, 55-65
```
ageorgou:
```suggestion
# groups 15-24, 25-34, 35-44, 45-54, 55-65
```

src/hivpy/sex_behaviour_data.py
```diff
@@ -0,0 +1,179 @@
...
+                                            [0.43,  0.3,  0.23,  0.03, 0.01],
+                                            [0.18,  0.18, 0.27,  0.27, 0.1]]])
+
+sex_mixing_matrix_female_options = np.array([[[0.43, 0.34, 0.12, 0.1,  0.01],
+                                              [0.09, 0.49, 0.3,  0.1,  0.02],
+                                              [0.03, 0.25, 0.34, 0.25, 0.13],
+                                              [0.0,  0.0,  0.05, 0.7,  0.25],
+                                              [0.0,  0.0,  0.0,  0.1,  0.9]]])
+
+# Doesn't seem to be interpolation between 10,15,20,25,30,35 on high?
```
ageorgou, line 167: No, I also think the original model only chooses from these specific values.

src/hivpy/sexual_behaviour.py
```diff
@@ -0,0 +1,48 @@
...
+
+from . import sex_behaviour_data as sb
+
+
+class SexualBehaviourModule:
+
+    def __init__(self, **kwargs):
+        # Randomly initialise sexual behaviour group transitions
+        self.sex_behaviour_trans_male = np.random.choice(sb.sex_behaviour_trans_male_options)
+        self.sex_behaviour_trans_female = np.random.choice(sb.sex_behaviour_trans_female_options)
```
ageorgou: Maybe combining these in one would simplify things a bit (here and when they're accessed)? For example, a dictionary where the key is male or female. Not sure though.
(and similarly for the other attributes)

src/hivpy/sexual_behaviour.py
```diff
@@ -0,0 +1,48 @@
...
+        if(gender == 1):
+            transition_matrix = self.sex_behaviour_trans_female
+            gender_index = 1
+        else:
+            transition_matrix = self.sex_behaviour_trans_male
+            gender_index = 0
+
+        age_index = min((int(age)-15)//5, 9)
+
+        risk_factor = self.baseline_risk[age_index][gender_index]
```
ageorgou: I'm missing where this comes into play in the original model at the moment but I'll keep looking

ageorgou (commented):
This makes sense so far. I've left some very preliminary comments which may be invalidated by how it goes on.

I'll have a think about the vectorization of the transition between sexual behaviour groups. At the very least we could do it separately for each sex, which would only call it twice overall, not once per individual. There may be a better way though.

src/hivpy/sexual_behaviour.py
mmcleod89: This is eq 1 on page 8 of the model details doc, but I'm not sure where it comes into the SAAS code. 
It seems to make the male version work but cause the female version to fail though, because the female matrices seem to already be normalised, so something definitely has to change but I'm not sure what yet.

src/hivpy/sex_behaviour_data.py
mmcleod89: We can presumably interpolate this at some point, but doesn't need to be done for the demo

src/hivpy/sexual_behaviour.py
mmcleod89: Agreed, it is clearer to have things be able to iterate over sex and group things as a single structure; have used dicts for each of these pairs for now.

src/hivpy/sex_behaviour_data.py
mmcleod89: See comment on the probabilities calculation below... not sure which is correct yet!

src/hivpy/sexual_behaviour.py
```diff
@@ -1,35 +1,67 @@
...
+                                  self.select_matrix(sb.sex_mixing_matrix_male_options),
+                                  SexType.Female:
+                                  self.select_matrix(sb.sex_mixing_matrix_female_options)}
+        self.short_term_partners = {SexType.Male:
+                                    self.select_matrix(sb.short_term_partners_male_options),
+                                    SexType.Female:
+                                    self.select_matrix(sb.short_term_partners_female_options)}
+
+    # Haven't been able to locate the probabilities for this yet
+    # Doing them uniform for now
```
ageorgou, line 52: It's not entirely straightforward. As far as I can see, the base weights for men are: 0.85, 0.10, 0.05, 0.003
and for women: 0.97, 0.03
But the weights beyond the first category are "personalised" by a risk factor (`rred`), which depends on age and other things (similar to what you have implemented for each time step, but a bit simpler).

Uniform is fine for now!

src/hivpy/sexual_behaviour.py
```diff
@@ -1,35 +1,67 @@
...
-            transition_matrix = self.sex_behaviour_trans_male
-            gender_index = 0
+        group i to group j, based on sex and age."""
+        transition_matrix = self.sex_behaviour_trans[sex]
 
-        age_index = min((int(age)-15)//5, 9)
+        age_index = np.minimum((age.astype(int)-15)//5, 9)
 
-        risk_factor = self.baseline_risk[age_index][gender_index]
+        risk_factor = (self.baseline_risk[age_index]).transpose()[sex.value]
```
ageorgou: I'd remove the `(...)` as I don't think they make it more readable, but that's just me!
```suggestion
        risk_factor = self.baseline_risk[age_index].transpose()[sex.value]
```
Actually, do we need the `transpose`? Why not index both dimensions at the same time:
```suggestion
        risk_factor = self.baseline_risk[age_index, sex.value]
```

ageorgou: And a note for the future: it would be nice to move consistently to using the `SexType` values directly rather than as numerical indices, but that can come when we read in the data from a static file.

src/hivpy/sexual_behaviour.py
```diff
@@ -40,9 +72,38 @@ def prob_transition(self, gender, age, i, j):
...
+                    for j in range(dim):
+                        Pmin = Pmax
+                        Pmax.loc[index] += self.prob_transition(sex, ages, i.value, j)
+                        population.loc[index & (rands >= Pmin) & (
+                            rands < Pmax), "sex_behaviour"] = j
+
+
+def selector(population, **kwargs):
+    index = reduce(operator.and_,
+                   (op(population[kw], val) for kw, (op, val) in kwargs.items()))
```
ageorgou, line 113: I like the generalisation of allowing different operators and think it will be useful. It does make the calls to this a bit more verbose and harder to parse. One way that might help would be to have helper funtions like `equals(value)`, `greater_than(value)` etc that return the relevant tuple. I think it's fine for now as it is, though.

src/hivpy/sexual_behaviour.py
```diff
@@ -40,9 +72,38 @@ def prob_transition(self, gender, age, i, j):
...
+                    self.short_term_partners[sex][g.value].rvs(size=sum(index)))
+
+    def update_sex_groups(self, population: pd.DataFrame):
+        """Determine changes to sexual behaviour groups.
+           Loops over sex, and behaviour groups within each sex.
+           Within each group it then loops over groups again to check all transition pairs (i,j)."""
+        pop_size = len(population)
+
+        for sex in SexType:
+            for i in SexBehaviours[sex]:
```
ageorgou: I would be explicit about this before we get lost with i and j, for example:

```suggestion
            for old_behaviour in SexBehaviours[sex]:
```

src/hivpy/sexual_behaviour.py
```diff
@@ -40,9 +72,38 @@ def prob_transition(self, gender, age, i, j):
...
+                index = selector(population, sex=(operator.eq, sex), sex_behaviour=(
+                    operator.eq, i.value), age=(operator.ge, 15))
+                rands = np.random.uniform(0.0, 1.0, pop_size)
+                ages = population.loc[index, "age"]
+                if(len(ages) > 0):
+                    dim = self.sex_behaviour_trans[sex].shape[0]
+                    Pmin = pd.Series([0.]*pop_size)
+                    Pmax = pd.Series([0.]*pop_size)
+                    for j in range(dim):
```
ageorgou: As above (again, for instance):
```suggestion
                    for new_behaviour in range(dim):
```

src/hivpy/sexual_behaviour.py
```diff
@@ -40,9 +72,38 @@ def prob_transition(self, gender, age, i, j):
...
+        for sex in SexType:
+            for i in SexBehaviours[sex]:
+                index = selector(population, sex=(operator.eq, sex), sex_behaviour=(
+                    operator.eq, i.value), age=(operator.ge, 15))
+                rands = np.random.uniform(0.0, 1.0, pop_size)
+                ages = population.loc[index, "age"]
+                if(len(ages) > 0):
+                    dim = self.sex_behaviour_trans[sex].shape[0]
+                    Pmin = pd.Series([0.]*pop_size)
+                    Pmax = pd.Series([0.]*pop_size)
```
ageorgou: Can we define these only for the people of concern instead and compare with `rands[index]` (if that's not the same size)? Though I don't expect the speedup will be that significant.

src/hivpy/sexual_behaviour.py
```diff
@@ -40,9 +72,38 @@ def prob_transition(self, gender, age, i, j):
...
+           Within each group it then loops over groups again to check all transition pairs (i,j)."""
+        pop_size = len(population)
+
+        for sex in SexType:
+            for i in SexBehaviours[sex]:
+                index = selector(population, sex=(operator.eq, sex), sex_behaviour=(
+                    operator.eq, i.value), age=(operator.ge, 15))
+                rands = np.random.uniform(0.0, 1.0, pop_size)
+                ages = population.loc[index, "age"]
+                if(len(ages) > 0):
```
ageorgou: Very minor, but this seems more natural to me (I stopped and wondered why ages would be empty if they're all  defined!)
```suggestion
                if any(index):
```

But actually, we can move this up a couple of lines so `rands` can be sampled for the relevant size too.

src/hivpy/demographics.py
```diff
@@ -177,7 +184,7 @@ def __init__(self, **kwargs):
     def initialize_sex(self, count):
         sex_distribution = (
             self.params['female_ratio'], 1 - self.params['female_ratio'])
-        return np.random.choice(SexType.categories, count, sex_distribution)
+        return pd.Series(np.random.choice(SexType, count, sex_distribution)).astype(SexDType)
```
ageorgou, line 187: So in the end was the performance hit from the categorical not significant? I lost track of where we left it!

src/tests/test_sexual_behaviour.py
```diff
@@ -1,22 +1,58 @@
...
-    assert (sex_behaviour_trans_male.shape == (15, 4, 4))
-    assert (sex_behaviour_trans_female.shape == (15, 2, 2))
-    check_prob_sums(1, sex_behaviour_trans_female)
-    check_prob_sums(0, sex_behaviour_trans_male)
+    for trans_matrix in sbd.sex_behaviour_trans_male_options:
+        assert (trans_matrix.shape == (4, 4))
+        check_prob_sums(SexType.Male, trans_matrix)
+    # for trans_matrix in sbd.sex_behaviour_trans_female_options:
+    #     assert (trans_matrix.shape == (2, 2))
+    #     check_prob_sums(1, trans_matrix)
```
ageorgou: The commented-out tests also pass for me (changing 1 to `SexType.Female`)!

src/hivpy/sexual_behaviour.py
```diff
@@ -1,35 +1,67 @@
...
-            transition_matrix = self.sex_behaviour_trans_female
-            gender_index = 1
-        else:
-            transition_matrix = self.sex_behaviour_trans_male
-            gender_index = 0
+        group i to group j, based on sex and age."""
+        transition_matrix = self.sex_behaviour_trans[sex]
 
-        age_index = min((int(age)-15)//5, 9)
+        age_index = np.minimum((age.astype(int)-15)//5, 9)
```
ageorgou: Am I right that the 9 here is because the `baseline_risk` matrix has 10 rows, to avoid indexing outside its bounds? If so, better avoid hardcoding it. Could have a field on the object that stores `self.baseline_risk.shape[1]` to avoid recomputing it.

ageorgou (commented):
Nice, this looks good. I have some small suggestions, but mostly for future changes. The only ones that really need addressed/clarified now are in `update_sex_groups` and the tests.

We can leave refinements for the future - I will open an issue to track these.

ageorgou:
The test failing is the one about deaths occurring, which is no longer valid. We can disable it for now, I'll adapt it while addressing #17.

src/hivpy/sexual_behaviour.py
mmcleod89: That is what I initially tried but I couldn't get it to work because of the need to combine the probabilistic condition with the index array (which has the same length as the full population) earlier:
`population.loc[index & (rands >= Pmin) & (rands < Pmax), "sex_behaviour"] = new_group`
There could be another way to do this but I couldn't find a good way ....

src/hivpy/sexual_behaviour.py
mmcleod89: That is better, I'm not that familiar with this `array[i,j]` notation which I will need to brush up on because I'm used to `array[i][j]` but for 2D arrays, but this doesn't work if your first index is a list of indices!

src/hivpy/sexual_behaviour.py
mmcleod89: Will update this

src/hivpy/sexual_behaviour.py
mmcleod89: Ditto

src/hivpy/demographics.py
mmcleod89: There was a bit of a performance hit, but it doesn't seem to be a big thing. `update_sex_groups` is also sex dependent but doesn't seem to be slowing things down very much, so I wonder if the extra time that it took was set-up time rather than performance time for using it when looking things up etc. 
I don't have strong feelings about using it either way though.

src/tests/test_sexual_behaviour.py
mmcleod89: I will put it back in properly fixed; it might just be that I broke the test while rewriting the functions!

ageorgou (approved):
Tests still failing because of no deaths, will be handled in different PR.

PR #22, opened 2022-01-26
Enhancement/hiv status

mmcleod89:
Calculate HIV status updates using simple probabilistic model

### DIFF
--- a/src/hivpy/hiv_status.py
+++ b/src/hivpy/hiv_status.py
@@ -1,0 +1,37 @@
+import operator
+
+import numpy as np
+import pandas as pd
+
+from .sexual_behaviour import selector
+
+
+class HIVStatusModule:
+    hiv_a = -0.00016  # placeholder value for now
+    hiv_b = 0.0128    # placeholder value for now
+    hiv_c = -0.156    # placeholder value for now
+
+    def _prob_HIV_initial(self, age):
+        """Completely arbitrary placeholder function for initial HIV presence in population"""
+        return np.clip(self.hiv_a*age**2 + self.hiv_b*age + self.hiv_c, 0.0, 1.0)
+
+    def initial_HIV_status(self, population: pd.DataFrame):
+        """Initialise HIV status based on age (& sex?)"""
+        """Assume zero prevalence for age < 15"""
+        hiv_status = self._prob_HIV_initial(population["age"]) > np.random.rand(len(population))
+        return hiv_status
+
+    def update_HIV_status(self, population: pd.DataFrame):
+        """Update HIV status for new transmissions in the last time period.\\
+            Super simple model where probability of being infected by a given person
+            is prevalence times transmission risk (P x r).\\
+            Probability of each new partner not infecting you then is (1-Pr)\\
+            Then prob of n partners independently not infecting you is (1-Pr)**n\\
+            So probability of infection is 1-((1-Pr)**n)"""
+        HIV_neg_idx = selector(population, HIV_status=(operator.eq, False))
+        rands = np.random.uniform(0.0, 1.0, sum(HIV_neg_idx))
+        HIV_prevalence = sum(population['HIV_status'])/len(population)
+        HIV_infection_risk = 0.005  # made up, based loosely on transmission probabilities
+        n_partners = population.loc[HIV_neg_idx, "num_partners"]
+        HIV_prob = 1-((1-HIV_prevalence*HIV_infection_risk)**n_partners)
+        population.loc[HIV_neg_idx, "HIV_status"] = (HIV_prob <= rands)
--- a/src/hivpy/hiv_status.py
+++ b/src/hivpy/hiv_status.py
@@ -34,4 +34,4 @@ class HIVStatusModule:
         HIV_infection_risk = 0.005  # made up, based loosely on transmission probabilities
         n_partners = population.loc[HIV_neg_idx, "num_partners"]
         HIV_prob = 1-((1-HIV_prevalence*HIV_infection_risk)**n_partners)
-        population.loc[HIV_neg_idx, "HIV_status"] = (HIV_prob <= rands)
+        population.loc[HIV_neg_idx, "HIV_status"] = (rands <= HIV_prob)
--- a/src/hivpy/sexual_behaviour.py
+++ b/src/hivpy/sexual_behaviour.py
@@ -103,7 +103,7 @@ class SexualBehaviourModule:
                     Pmin = np.zeros(subpop_size)
                     Pmax = np.zeros(subpop_size)
                     for new_group in range(dim):
-                        Pmin = Pmax
+                        Pmin = Pmax.copy()
                         Pmax += self.prob_transition(sex, ages, prev_group, new_group)
                         # This has to be a Series so it can be combined with index correctly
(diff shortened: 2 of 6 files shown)


src/hivpy/hiv_status.py
```diff
@@ -0,0 +1,37 @@
...
+            Probability of each new partner not infecting you then is (1-Pr)\\
+            Then prob of n partners independently not infecting you is (1-Pr)**n\\
+            So probability of infection is 1-((1-Pr)**n)"""
+        HIV_neg_idx = selector(population, HIV_status=(operator.eq, False))
+        rands = np.random.uniform(0.0, 1.0, sum(HIV_neg_idx))
+        HIV_prevalence = sum(population['HIV_status'])/len(population)
+        HIV_infection_risk = 0.005  # made up, based loosely on transmission probabilities
+        n_partners = population.loc[HIV_neg_idx, "num_partners"]
+        HIV_prob = 1-((1-HIV_prevalence*HIV_infection_risk)**n_partners)
+        population.loc[HIV_neg_idx, "HIV_status"] = (HIV_prob <= rands)
```
ageorgou: I think the comparison is the other way around?

src/hivpy/hiv_status.py
```diff
@@ -0,0 +1,37 @@
...
+    hiv_b = 0.0128    # placeholder value for now
+    hiv_c = -0.156    # placeholder value for now
+
+    def _prob_HIV_initial(self, age):
+        """Completely arbitrary placeholder function for initial HIV presence in population"""
+        return np.clip(self.hiv_a*age**2 + self.hiv_b*age + self.hiv_c, 0.0, 1.0)
+
+    def initial_HIV_status(self, population: pd.DataFrame):
+        """Initialise HIV status based on age (& sex?)"""
+        """Assume zero prevalence for age < 15"""
```
ageorgou, line 20: We don't enforce this currently, do we? Or does the placeholder distribution ensure that?

ageorgou (changes requested):
Looks fine for a simplified initial version, apart from the comparison (I think). Could add some sanity tests, e.g. that someone who is HIV+ does not become negative subesquently.

ageorgou (approved):
All comments addressed, tests still failing because of deaths, will be fixed in separate PR.

src/hivpy/sexual_behaviour.py
```diff
@@ -51,9 +51,11 @@ def __init__(self, **kwargs):
...
     # Doing them uniform for now
     def init_sex_behaviour_groups(self, population):
-        population["sex_behaviour"] = np.where(population['sex'] == SexType.Male,
-                                               np.random.choice(MaleSexBehaviour).value,
-                                               np.random.choice(FemaleSexBehaviour).value)
+        for sex in SexType:
+            index = selector(population, sex=(operator.eq, sex))
+            n_sex = sum(index)
+            population.loc[index, "sex_behaviour"] = np.random.choice(
+                SexBehaviours[sex], size=n_sex)
```
ageorgou, line 58: Out of curiosity, is this so we only generate the values we need or was there something else wrong in the old version?

src/hivpy/sexual_behaviour.py
```diff
@@ -100,8 +103,8 @@ def update_sex_groups(self, population: pd.DataFrame):
                     Pmin = np.zeros(subpop_size)
                     Pmax = np.zeros(subpop_size)
                     for new_group in range(dim):
-                        Pmin = Pmax
-                        Pmax += self.prob_transition(sex, ages, prev_group.value, new_group)
+                        Pmin = Pmax.copy()
```
ageorgou, line 106: Good catch. We'll need to be aware of this, as well as checking whether something returns a view into a dataframe or a copy.

src/tests/test_hiv_status.py
```diff
@@ -22,16 +22,18 @@ def test_initial_hiv():
...
-    HIV_module = HIVStatusModule()
-    pop["HIV_status"] = HIV_module.initial_HIV_status(pop)
-    # to keep this independent just creating partners for everyone
+    pop = Population(size=pop_size, start_date=date(1989, 1, 1))
+    data = pop.data
+    prev_status = data["HIV_status"].copy()
+
     for i in range(10):
-        prev_status = pop["HIV_status"].copy()
-        pop["num_partners"] = pd.Series([2]*pop_size)
```
ageorgou: Just double checking that this was left out on purpose (so the test is still meaningful without setting partners for everyone)

src/tests/test_sexual_behaviour.py
```diff
@@ -33,6 +35,9 @@ def check_num_partners(row):
     sex = row["sex"]
     group = row["sex_behaviour"]
     n = row["num_partners"]
+    age = row["age"]
+    if age <= 15:
+        return n == 0
```
ageorgou: I would maybe add a comment here to explain why this is needed, for future readers.

src/tests/test_sexual_behaviour.py
```diff
@@ -52,7 +57,25 @@ def check_num_partners(row):
...
     assert np.all(checks)
+
+
+def test_sex_behaviour_groupings():
+    """Check that people are assigned to all sex behaviour groups!"""
+    pop_data = Population(size=1000, start_date=date(1989, 1, 1)).data
+    for sex in SexType:
+        for group in SexBehaviours[sex]:
+            index = selector(pop_data, sex=(operator.eq, sex), sex_behaviour=(operator.eq, group))
+            assert any(index)
```
ageorgou: This is fine, although one suggestion to also check that we are not assigning values outside the expected (I think in that case pandas will silently change the type of the column instead of complaining, but I may be wrong):
```suggestion
    for sex in SexType:
        groups_in_data = pop_data[pop_data.sex == sex].sex_behaviour.unique()
        assert sorted(groups_in_data) == sorted(SexBehaviours[sex])
```

src/tests/test_sexual_behaviour.py
```diff
@@ -52,7 +57,25 @@ def check_num_partners(row):
...
+
+
+def test_behaviour_updates():
```
ageorgou, line 72: Could add a docstring to clarify the intent here, and possibly update it if the test evolves.

```suggestion
def test_behaviour_updates():
    """Check that at least one behaviour group changes with time."""
```

ageorgou (approved):
New changes look good, only some minor questions and suggestions, but happy to merge this.

src/tests/test_hiv_status.py
mmcleod89: Yes, I actually found that it was better to use the ordinary initialisation that the population object will do for now as it prevent people from being assigned partners if they are under 15 / not born yet. So `num_partners` is still set here, but this is no longer independent of the function that generates the number of partners. We could recreate this independence by using a better assignment in the test that takes into account things like the ages but for now checking that it worked with the real thing was useful.

src/hivpy/sexual_behaviour.py
mmcleod89: The old version wasn't assigning a unique group for each person, but for each gender would make one random choice out of their groups and assign that to everyone of that gender! Essentially creating an array of copies of a random choice, rather than an array of random choices.

PR #26, opened 2022-01-31
Enhancement/simulation output

mmcleod89:
Also amped up the HIV transmission parameters to check that transmissions were happening properly and get more changeable outputs as we had almost no transmissions with a population of 10,000 and only about 10 transmissions over a simulation time of 200 steps with 100,000 people. 

Individual simulation output files are labelled with a datetime to distinguish them if multiple simulations are run, given that we don't have any other simulation ID implemented at the moment. 

Output directory and filename should come from the config but I need to figure out how the simluationConfig works first


src/hivpy/simulation.py
```diff
@@ -1,10 +1,87 @@
...
+        men_idx = selector(pop_data, sex=(operator.eq, SexType.Male))
+        self.output_stats["HIV prevalence (male)"][self.step] = (
+            self._ratio(HIV_pos_idx & men_idx, men_idx & over_15_idx))
+        women_idx = selector(pop_data, sex=(operator.eq, SexType.Female))
+        self.output_stats["HIV prevalence (female)"][self.step] = self._ratio(
+            HIV_pos_idx & women_idx, women_idx & over_15_idx)
+
+        # Update HIV prevalence by age
+        for age_bound in range(self.age_min, self.age_max, self.age_step):
+            key = "HIV prevalence (" + str(age_bound) + "-" + str(age_bound+(self.age_step-1)) + ")"
```
ageorgou:
```suggestion
            key = f"HIV prevalence ({age_bound}-{age_bound+(self.age_step-1)})"
```

ageorgou (approved):

PR #27, opened 2022-02-01
Add death rate

ageorgou:
Addressing #17.

This adds sex- and age-related death rates based on the SAS model (but not the exact numbers yet).
Still pending:
- [x] Checking the performance of the age selection and possibly replacing it
- [ ] Reading in values from a file
- [ ] Tests on death rate


src/hivpy/demographics.py
```diff
@@ -193,3 +229,27 @@ def initialise_age(self, count):
...
+                index = selector(population_data, sex=(operator.eq, sex),
+                                 age=(between, (low_age, high_age)))
+                group = population_data[index]
+                # Probability of dying, assuming time step of 3 months
+                prob_of_death = 1 - exp(-rate / 4)
+                died = np.random.choice([True, False], size=len(group),
+                                        p=[prob_of_death, 1 - prob_of_death])
+                # Mark those who died from this group
+                all_died |= pd.Series(died, index=group.index)
+        return population_data.date_of_death.isnull() & all_died
```
mmcleod89, line 252: Why not incorporate the condition that they be not already dead into the selector so we don't need to calculate the probabilities for dead people?

src/hivpy/demographics.py
```diff
@@ -193,3 +229,27 @@ def initialise_age(self, count):
...
+        """Get which individuals die in a time step, as a boolean Series."""
+        all_died = pd.Series([False] * len(population_data))
+        for sex in SexType:
+            for (low_age, high_age), rate in self.params["death_rates"][sex].items():
+                # Find everyone in the relevant age range
+                index = selector(population_data, sex=(operator.eq, sex),
+                                 age=(between, (low_age, high_age)))
+                group = population_data[index]
+                # Probability of dying, assuming time step of 3 months
+                prob_of_death = 1 - exp(-rate / 4)
```
mmcleod89, line 247: Can you explain this probability? This doesn't look correct to me

src/tests/test_demographics.py
```diff
@@ -59,3 +61,27 @@ def test_stepwise_age_distribution(stepwise_age_module):
...
+    """Check that that death is not recorded again in future steps."""
+    population_data = pd.DataFrame({
+        'date_of_death': [None, datetime.today(), datetime.today() - timedelta(days=1)],
+        'age': [30, 20, 50],
+        'sex': [SexType.Female, SexType.Female, SexType.Male]
+    })
+    new_deaths = default_module.determine_deaths(population_data)
+    # Find who already had a date of death, and check that they are not marked
+    # as having died in this time step.
+    assert not new_deaths[population_data.date_of_death.notnull()].any()
```
mmcleod89, line 76: Surely this always passes because no one is dead when the population is created, and the population isn't evolved before calculating `new_deaths`, so `population_data.date_of_death.notnull()` would always be empty?

src/tests/test_simulation.py
```diff
@@ -78,4 +78,5 @@ def test_death_occurs(tmp_path):
     # Check that the number alive never grows... (some steps may have 0 deaths)
     assert all(results.num_alive.diff()[1:] <= 0)
     # ...and that there is at least one death overall!
-    assert results.num_alive[-1] < results.num_alive[0]
+    # FIXME This is not guaranteed at the moment because of the values used
+    # assert results.num_alive[-1] < results.num_alive[0]
```
mmcleod89, line 82: This isn't really about your pull request here, but this bit of code actually raises an interesting point - shouldn't the number of people alive grow? Because the population is growing as people are born? Are we counting the unborn as alive here?

src/tests/test_demographics.py
ageorgou: This isn't using a `Population`, but rather the specific data frame defined here, which has two people marked as dead already.

src/tests/test_simulation.py
ageorgou: That's a good point. Looking at it, the calculation does include people who are not born, possibly because when it was written everyone was born.

It's using the old "tracked attributes", so we can change this as part of replacing those with `SimulationOutcome`s. Though I'm not sure it's guaranteed to be consistently growing if deaths exceed births at some time step. Perhaps that's unlikely enough to happen with the specific configuration we have though.

src/hivpy/demographics.py
ageorgou: I was copying the relevant formula from the old model (involving `ac_death_rate`). The division by 4 is assuming a 3 month interval, because the overall death rate refers to a whole year. I think I tried to derive it but wasn't sure exactly what "rate" this was, so I left it as-is. Worth clarifying for sure, I'll check again in the model description.

mmcleod89: Okay I have tried the calculation out and I think that this is correct if the original probability for a year is `1-exp(-rate)`, based ensuring that the probability of dying in a year is the same as the sum of the probabilities of:

- dying in the first period
- surviving first period and dying in second
- surviving first two and dying in third
- etc.

then we have basically a constant (probability of dying in a period) multiplying a geometric sum (probability of surviving a period to the power of the number of periods they survive), which ends up giving me a formula for probability of dying in a year divided by N:
P_N = 1 - (1 - P_year)^(1/N)
which all works out fine.

src/tests/test_simulation.py
mmcleod89: Yeah it's definitely not guaranteed to be consistently growing, but we would expect it to usually be growing (since populations are trending that way in general) and definitely not expect it to be consistently declining, except when we count the unborn. But I agree we can revisit this when we address simulation outputs.

src/hivpy/demographics.py
```diff
@@ -193,3 +229,27 @@ def initialise_age(self, count):
...
+                # Mark those who died from this group
+                all_died |= pd.Series(died, index=group.index)
+        return population_data.date_of_death.isnull() & all_died
+
+        # rates = np.ones(len(population_data)) * self.params["death_rate"]
+        # Assuming time step of 3 months
+        # probs = np.exp(-rates / 4)
+        probs = 1 - np.exp(-self.params["death_rate"] / 4)
+        return (population_data.date_of_death.isnull()
+                & np.random.choice([True, False], size=len(population_data), p=[probs, 1-probs]))
```
mmcleod89: I didn't notice this earlier but this code is after the return statement and never executed, and appears to be an older version of this function, so we should probably cut this!

ageorgou:
Will tackle more tests and reading in from file in a separate PR, to avoid this and #35 becoming too complex and interlinked.

src/hivpy/demographics.py
```diff
@@ -246,3 +250,21 @@ def determine_deaths(self, population_data: pd.DataFrame) -> pd.Series:
...
+    def determine_deaths(self, population_data: pd.DataFrame) -> pd.Series:
+        """Get which individuals die in a time step, as a boolean Series."""
+        # This binning should perhaps happen when the date advances
+        # Age groups are the same regardless of sex
+        age_limits = [lower for (lower, upper) in DEATH_RATE_MALE] + [inf]
+        population_data["age_group"] = np.digitize(population_data.age, age_limits)
+
+        death_probs = np.full(len(population_data), np.nan)
+        groups = population_data.groupby(["sex", "age_group"])
+        for (sex, age_group), entries in groups.groups.items():
```
mmcleod89: maybe rename `groups` so we don't have the slightly confusing `groups.groups.items()`

mmcleod89 (approved):

PR #34, opened 2022-02-10
Add initial sexual behaviour groups  & test; add output/log to gitignore

### LINKED ISSUE #30: Initial distribution of number of sexual behaviour groups

mmcleod89:


src/hivpy/sex_behaviour_data.py
```diff
@@ -1,6 +1,19 @@
 import numpy as np
 import scipy.stats as stat
 
+from .demographics import SexType
+
+# Initial sexual behaviour group probabilities
+# Taken from the SAS code lines 1249 and 1283
```
ageorgou:
```suggestion
# Taken from the SAS code Section 2
```
(who knows when line numbers will change!)

src/hivpy/sexual_behaviour.py
```diff
@@ -29,12 +29,18 @@ class SexualBehaviourModule:
     def select_matrix(self, matrix_list):
         return matrix_list[np.random.choice(matrix_list.shape[0])]
 
+    def norm_probs(self, prob_dict: dict):
+        for key, data in prob_dict.items():
+            data /= sum(data)
```
ageorgou: Possibly paranoid here but I'm slightly concerned that `norm_probs` updates in place. This works with numpy arrays (or Python lists) because iterating over them works "by reference". If we switch to something else, though (or if numpy changes the implementation, which I think has happened in the past), this might break. Is it better to return a normalised version instead?
```suggestion
    def norm_probs(self, prob_dict: dict):
        return {
            key: data / sum(data)
            for key, data in prob_dict.items()
        }
```
But not convinced it is.

ageorgou: This could also be renamed `_norm_probs` if we want to stress that it should be thought of as private.

src/tests/test_sexual_behaviour.py
```diff
@@ -77,3 +79,21 @@ def test_behaviour_updates():
...
+        n_sex = sum(index_sex)
+        Prob_sex = probs[sex].copy()
+        Prob_sex /= sum(Prob_sex)
+        for g in SexBehaviours[sex]:
+            index_group = selector(pop_data, sex_behaviour=(operator.eq, g))
+            p = Prob_sex[g]
+            sigma = np.sqrt(p*(1-p)*float(n_sex))
+            E = float(n_sex)*p
+            N_g = sum(index_sex & index_group)
+            assert(((E - sigma*3) <= N_g) and (N_g <= (E + sigma*5)))
```
ageorgou: Scalar numbers have convenient double-sided comparison!
```suggestion
            assert E - sigma*3 <= N_g <= E + sigma*5
```

(Also, I'd avoid parentheses after `assert`. You can end up asserting a tuple accidentally, which will always be "truthy" even if it contains only `False` values)

src/tests/test_sexual_behaviour.py
```diff
@@ -77,3 +79,21 @@ def test_behaviour_updates():
         pop.sexual_behaviour.update_sex_groups(pop.data)
     subsequent_groupings = pop.data["sex_behaviour"]
     assert(any(initial_groupings != subsequent_groupings))
+
+
+def test_initial_sex_behaviour_groups():
```
ageorgou, line 76: This makes sense but, I think, could be a bit clearer.
I think it'd be very helpful to have a docstring here explaining exactly the aim of the test (that initial distribution matches expected one within a tolerance of ...). Failing that (or even so!), rename some variables or add a comment with their meaning, especially E (even though the symbol matches the expectation - no pun intended).

src/tests/test_sexual_behaviour.py
ageorgou: Separately, is there a reason the interval is not symmetric?

ageorgou (approved):
Looks good, minor suggestions and questions.

src/hivpy/sex_behaviour_data.py
mmcleod89: I did think about this, and happy to change that over. I'm only intending for this to be a temporary comment as at some point either the model doc - or hivpy itself - will become the authority!

src/hivpy/sexual_behaviour.py
mmcleod89: This is true; I had a think about this too and wasn't sure which was the best way to go; I was mostly concerned about a) doing an unnecessary deep copy and b) having the version of the array in `sexual_behaviour_data` differing from the one that's in the `sexual_behaviour` module (un-normalised and normalised respectively). The `data` version shouldn't really be accessed from that point on though, but it does still exist and could easily be referenced accidentally. That said, I think that this will change anyway once we're loading things from yaml files, as then it could just be normalised at load time and we won't have two copies floating around.

src/tests/test_sexual_behaviour.py
mmcleod89: There is no reason the interval is not symmetric, I think that was just a typo. (I initially set *5 but thought that this was a bit too lenient as a test, and went down to *3, but evidently something went awry!)

PR #35, opened 2022-03-28
Enhancement/sex behaviour data

mmcleod89:


mmcleod89:
I can't find a way of reducing the line lengths on these files because of all the long names for data fields so I keep failing static checks

src/hivpy/sex_behaviour_data.py
```diff
@@ -17,7 +18,10 @@ def _get_discrete_dist(self, keys):
         dist_data = self.data
         for k in keys:
             dist_data = dist_data[k]
-        return stat.rv_discrete(values=(dist_data["Value"], dist_data["Probs"]))
+        values = np.array(dist_data["Value"])
+        probs = np.array(dist_data["Probs"])
+        probs /= sum(probs)
```
ageorgou, line 36: Should this normalisation need to happen in `_get_discrete_dist_list` too? Or is it only the initial probabilities that are unnormalised?

src/hivpy/sex_behaviour_data.py
```diff
@@ -30,14 +34,23 @@ def _norm_probs(self, prob_dict: dict):
             for key, data in prob_dict.items()
         }
 
+    def _select_matrix(self, matrix_list):
+        return matrix_list[np.random.choice(len(matrix_list))]
```
ageorgou, line 51: This would also work and appear more direct, unless I'm missing a dimension or something?
```suggestion
        return np.random.choice(matrix_list)
```

src/tests/test_sexual_behaviour.py
```diff
@@ -78,11 +81,14 @@ def test_initial_sex_behaviour_groups():
...
-    probs = sbd.init_sex_behaviour
+    with open("data/sex_behaviour.yaml", 'r') as file:
+        yaml_data = yaml.safe_load(file)
+    probs = {SexType.Male: yaml_data["initial_sex_behaviour_probabilities"]["Male"]["Probs"],
+             SexType.Female: yaml_data["initial_sex_behaviour_probabilities"]["Female"]["Probs"]}
     for sex in SexType:
         index_sex = selector(pop_data, sex=(operator.eq, sex))
         n_sex = sum(index_sex)
-        Prob_sex = probs[sex].copy()
+        Prob_sex = np.array(probs[sex].copy())
```
ageorgou: Very minor: we shouldn't need to copy now that we're creating a new numpy array each time anyway.

src/tests/test_sexual_behaviour.py
```diff
@@ -23,10 +24,12 @@ def check_prob_sums(sex, trans_matrix):
 
 
 def test_transition_probabilities():
-    for trans_matrix in sbd.sex_behaviour_trans_male_options:
+    with open("data/sex_behaviour.yaml",'r') as file:
+        yaml_data = yaml.safe_load(file)
```
ageorgou: Should we test the data files themselves or the object? I'm leaning towards the latter, so that we also test (and stay independent of) how the files are read in `SexualBehaviourData`. That may also save us some normalisation etc in the tests? We could include the initialisation of the object in the test, or have it as a fixture if it's going to be used in more tests.

More of a philosophical question. Maybe something to consider as a future refinement.

ageorgou:
So this addresses #28 and #29, right?

src/hivpy/common.py
```diff
@@ -0,0 +1,45 @@
...
+        N = len(vals)
+        if(len(probs) != N):
+            raise Exception
+        index_range = np.arange(0, N, 1)
+        self.data = vals
+        self.dist = stat.rv_discrete(values=(index_range, probs))
+
+    def sample(self, size=1):
+        indices = self.dist.rvs(size=size)
+        return self.data[indices]
```
ageorgou, line 22: I'm a bit confused here. Why can't we use `rv_discrete` directly with the specific `vals`? If we really want the method to be called `sample`, we could wrap it in a new class, but converting to indices and looking up the corresponding values seems unnecessary.

src/hivpy/sex_behaviour_data.py
```diff
@@ -1,195 +1,96 @@
...
+
+class SexualBehaviourData:
+    """Class to hold and interpret sexual behaviour data loaded from the yaml file"""
+
+    def _setup_probabilty_dist(self, prob_dict):
+        if("Range" in prob_dict):
+            min = prob_dict["Range"][0]
+            max = prob_dict["Range"][1]
+            N = max - min + 1
+            return DiscreteChoice(np.arange(min, max+1, 1), np.array([1./N]*N))
```
ageorgou, line 17: If we end up using `rv_discrete` directly, this could be simplified to use [`randint`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.randint.html).

src/hivpy/sex_behaviour_data.py
```diff
@@ -1,195 +1,96 @@
...
+
+            self.rred_initial = self.data["rred_initial"]
+
+            self.new_partner_dist = self._get_discrete_dist("new_partner_factor")
+
+            self.p_rred_p_dist = self._get_discrete_dist("population_rred_personal")
+
+        except KeyError as ke:
+            print(ke.args)
+            raise ke
```
ageorgou: Should we throw the new `DataLoadException` here?

src/hivpy/sexual_behaviour.py
```diff
@@ -27,61 +26,78 @@ class FemaleSexBehaviour(IntEnum):
...
 
     def select_matrix(self, matrix_list):
-        return matrix_list[np.random.choice(matrix_list.shape[0])]
-
-    def _norm_probs(self, prob_dict: dict):
-        return {
-            key: data / sum(data)
-            for key, data in prob_dict.items()
-        }
+        return matrix_list[np.random.choice(len(matrix_list))]
```
ageorgou, line 29: I may have suggested this before and forgotten the reply, sorry. 
```suggestion
        return np.random.choice(matrix_list)
```

src/hivpy/sexual_behaviour.py
```diff
@@ -105,20 +121,14 @@ def update_sex_groups(self, population: pd.DataFrame):
...
                     Pmin = np.zeros(subpop_size)
                     Pmax = np.zeros(subpop_size)
                     for new_group in range(dim):
                         Pmin = Pmax.copy()
-                        Pmax += self.prob_transition(sex, ages, prev_group, new_group)
+                        Pmax += self.prob_transition(sex, rred, prev_group, new_group)
                         # This has to be a Series so it can be combined with index correctly
-                        jump_to_new_group = pd.Series((rands >= Pmin) & (rands < Pmax),
-                                                      index=ages.index)
+                        jump_to_new_group = pd.Series((rands >= Pmin) & (rands < Pmax))
```
ageorgou: Double checking that it's safe to remove the index from this Series here. Will it always be combined with `index` correctly? I remember there was  a reason for adding it but maybe it's no longer relevant in this version.

src/tests/test_sexual_behaviour.py
```diff
@@ -16,17 +17,26 @@ def check_prob_sums(sex, trans_matrix):
     (dim, _) = trans_matrix.shape
     for i in range(0, dim):
         ages = np.array([15, 20, 25, 30, 35, 40, 45, 50, 55, 60, 65])
+        sexes = np.array([sex]*len(ages))
+        pop = pd.DataFrame({"age": ages, "sex": sexes})
+        SBM.init_risk_factors(pop)
+        assert len(pop["rred"]) == 11
+        print(pop["new_partner_factor"])
+        for r in pop["new_partner_factor"]:
+            assert 0 < r <= 2
```
ageorgou: Minor, personal preference:
```suggestion
        assert all((0 < pop["new_partner_factor"]) & (pop["new_partner_factor"] <= 2))
```

src/tests/test_sexual_behaviour.py
```diff
@@ -16,17 +17,26 @@ def check_prob_sums(sex, trans_matrix):
...
         for j in range(0, dim):
-            tot_prob += SBM.prob_transition(sex, ages, i, j)
+            tot_prob += SBM.prob_transition(sex, pop["rred"], i, j)
         assert(np.allclose(tot_prob, 1.0))
 
 
 def test_transition_probabilities():
-    for trans_matrix in sbd.sex_behaviour_trans_male_options:
+    with open("data/sex_behaviour.yaml", 'r') as file:
+        yaml_data = yaml.safe_load(file)
```
ageorgou, line 33: Should we be creating a `SexualBehaviourData` object and checking on that?

src/hivpy/sex_behaviour_data.py
mmcleod89: The list of matrices gets loaded into a 3D array (this is done by py-yaml), but np.random.choice will only accept 1D arrays (it does not interpret this as an array of 2D arrays) e.g.:
```
>>> x = np.array([[[1.0,0.0,0.0],[0.0,1.0,0.0],[0.0,0.0,1.0]],
...               [[1.0,1.0,1.0],[0.0,0.0,0.0],[0.0,0.0,0.0]],
...               [[0.0,0.0,0.0],[1.0,1.0,1.0],[0.0,0.0,0.0]]])
>>> np.random.choice(x)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mtrand.pyx", line 911, in numpy.random.mtrand.RandomState.choice
ValueError: a must be 1-dimensional
```

So this is kind of a fudge to get around that without altering the yaml import or moving the data into an intermediate structure.

src/tests/test_sexual_behaviour.py
mmcleod89: That's true, I think the reason I tested the file instead is because this problem is arguably independent of the object - it's checking for invalid data in the input file, and so if it failed because there was a problem with, for example, the object failing to initialise properly, then that's kind of a separate issue. Might be worth looking at again in the future, although I think this test will become redundant as the probabilities will get internally renormalised.

src/hivpy/common.py
mmcleod89: Because the values in `rv_discrete` **must** be integers, which is extremely limiting. (Typically, we are sampling values which are floats.)

mmcleod89: It seems like all of these random choice / sampling methods have irritating caveats like this, which was another reason I decided a wrapper would be useful - to make it easier to swap out the back end implementation if we find something better without changing loads of code

src/tests/test_sexual_behaviour.py
mmcleod89: Same as above:

>The reason I tested the file instead is because this problem is arguably independent of the object - it's checking for invalid data in the input file, and so if it failed because there was a problem with, for example, the object failing to initialise properly, then that's kind of a separate issue. Might be worth looking at again in the future, although I think this test will become redundant as the probabilities will get internally renormalised.

src/hivpy/sexual_behaviour.py
mmcleod89: As above; causes error.

src/hivpy/common.py
ageorgou: Huh, that's odd, I completely missed that even when trying it out... That makes sense then.

src/hivpy/sex_behaviour_data.py
ageorgou: Okay, makes sense. I thought this was loaded as a list, not a 3D numpy array. The `random.choice` function from the standard library would work but it may be slower, and we'd have to keep track of multiple RNGs then.

ageorgou (approved):
Not sure why GitHub was showing conflicts previously, as could merge locally without issue.

From a quick look at [this discussion](https://github.community/t/merge-conflicts-on-github-not-locally/949/21), it seems that GitHub checks for conflicts at intermediate points rather than the tip of the branch? Oh well!

Issue #39, opened 2022-04-07
Update numpy sampling style


pweschmidt:
Use the latest style (the current one is deprecated)

PR #43, opened 2022-05-05
Switch to newer-style RNG calls

ageorgou:
Addressing #39.

- Move away from the legacy, shared-state `np.random.*` functions to the newer `Generator` API.
- This also lets us simplify some calls that were [previously discussed](https://github.com/UCL/hivpy/pull/35#discussion_r839645794). Is it still worth keeping the `select_matrix` function in `SexualBehaviourModule`?
- Also fix a bug with the distribution of sexes that became apparent while making the changes.

With the new format, everything shares the same `Generator`, which is initialised and imported from `hivpy.common`. One downside is that we can't set the seed as easily as before, although in the unit tests we can get around that by monkey-patching the shared `rng` object. Another option is to have a very simple wrapper around the generator, with a method for (re)setting the seed. At the moment we never use this in the tests, so I wasn't sure whether to add that in this PR or only later when we need it.

### DIFF
--- a/src/tests/test_demographics.py
+++ b/src/tests/test_demographics.py
@@ -25,7 +25,7 @@ def test_sex_distribution(default_module):
     count = 100000
     sex = default_module.initialize_sex(count)
     female = np.sum(sex == SexType.Female)
-    assert pytest.approx(female/count, rel=0.05) == FEMALE_RATIO
+    assert pytest.approx(female/count, rel=0.01) == FEMALE_RATIO
 
 
 def test_continuous_age_distribution(default_module):
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -47,7 +47,20 @@ def between(values, limits):
 
 
 class ResettableRandomState:
+    """A convenience class for using the NumPy random number generator.
+
+    This is meant to be used as: `from hivpy.common import rng`
+
+    For most things, this can be used exactly as NumPy's `Generator`. All the
+    methods like `random`, `normal`, `choice` are supported and called in
+    exactly the same way.
+
+    The primary purpose of this is to share the `Generator` across multiple files,
+    while allowing anyone to set the seed, whether in tests or the main code.
+    It also offers the ability to use a temporary seed.
+    """
     def __init__(self):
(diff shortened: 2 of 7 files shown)


src/tests/test_demographics.py
```diff
@@ -25,7 +25,7 @@ def test_sex_distribution(default_module):
     count = 100000
     sex = default_module.initialize_sex(count)
     female = np.sum(sex == SexType.Female)
-    assert pytest.approx(female/count, rel=0.05) == FEMALE_RATIO
+    assert pytest.approx(female/count, rel=0.01) == FEMALE_RATIO
```
mmcleod89, line 28: I feel like we should decide as a group on a standard for deciding the thresholds of probabilistic tests? We seem to go back and forth on these quite a bit so it would be good to be consistent. If we can codify this somewhere (in some shared variable for example) that would be even better...

mmcleod89 (commented):
In general happy with all this; I do think that we should remove `select_matrix` functions now that they just wrap the new `choice` function, taking the exact same argument without adding any other functionality or simplifying the interface. 

I do however think that we should implement the seeding functionality now if possible, given that it seems sensible to introduce it at the same time that we introduce the RNG (we don't know what future PR it will fit into), and this is functionality that we know that we will need later that we don't want to forget about. If it requires a lot of thought or moving things around though we can make it its own pull request, but I do think sooner is better than later.

src/tests/test_demographics.py
ageorgou: That would be good and give some more reassurance that we are not just enabling the tests to pass. Can we come up with a single rule, given that we have several distributions and generative process?

ageorgou:
I agree it's reasonable to add the seed setting here. I'll work on it and ask for another review, the solution I have in mind should work without moving much around.

mmcleod89 (approved):
I really like the temporary seed idea! Do we want to add a test / condition to check that the temporary seed is generating different random numbers, or is that superfluous? 
Happy for this to go through though!

src/tests/test_demographics.py
ageorgou: Followed up in #45.

PR #46, opened 2022-05-13
Enhancement/risk reduction

mmcleod89:
Still needs some additional tests to check that things are working as expected, although I'm not quite sure what all of those will look like.

### DIFF
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -18,8 +18,12 @@ class DiscreteChoice:
         self.dist = stat.rv_discrete(values=(index_range, probs))
 
     def sample(self, size=1):
-        indices = self.dist.rvs(size=size)
-        return self.data[indices]
+        if(size == 1):
+            index = self.dist.rvs()
+            return self.data[index]
+        else:
+            indices = self.dist.rvs(size=size)
+            return self.data[indices]
 
 
 class SexType(IntEnum):
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -18,13 +18,16 @@ class DiscreteChoice:
         self.data = vals
         self.dist = stat.rv_discrete(values=(index_range, probs))
 
-    def sample(self, size=1):
-        if(size == 1):
+    def sample(self, **kwargs):
+        """Samples from a random discrete distribution.
+        Used without an argument it will return a single object.
+        Used with the "size" argument it will return an array of that size."""
+        if "size" in kwargs:
+            indices = self.dist.rvs(size=kwargs["size"])
+            return self.data[indices]
+        else:
             index = self.dist.rvs()
             return self.data[index]
-        else:
-            indices = self.dist.rvs(size=size)
-            return self.data[indices]
 
 
 class SexType(IntEnum):
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -22,7 +22,7 @@ class DiscreteChoice:
         """Samples from a random discrete distribution.
         Used without an argument it will return a single object.
         Used with the "size" argument it will return an array of that size."""
-        if "size" is None:
+        if size is None:
             indices = self.dist.rvs()
             return self.data[indices]
         else:
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -18,15 +18,15 @@ class DiscreteChoice:
         self.data = vals
         self.dist = stat.rv_discrete(values=(index_range, probs))
 
-    def sample(self, **kwargs):
+    def sample(self, size=None):
         """Samples from a random discrete distribution.
         Used without an argument it will return a single object.
         Used with the "size" argument it will return an array of that size."""
-        if "size" in kwargs:
-            indices = self.dist.rvs(size=kwargs["size"])
+        if "size" is None:
+            indices = self.dist.rvs()
             return self.data[indices]
         else:
-            index = self.dist.rvs()
+            index = self.dist.rvs(size=size)
             return self.data[index]
 
 
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -10,7 +10,7 @@ import scipy.stats as stat
 
 
 class DiscreteChoice:
-    def __init__(self, vals, probs):
+    def __init__(self, vals: np.ndarray, probs):
         N = len(vals)
         if(len(probs) != N):
             raise Exception
@@ -23,11 +23,11 @@ class DiscreteChoice:
         Used without an argument it will return a single object.
         Used with the "size" argument it will return an array of that size."""
         if size is None:
-            indices = self.dist.rvs()
-            return self.data[indices]
-        else:
-            index = self.dist.rvs(size=size)
+            index = self.dist.rvs()
             return self.data[index]
+        else:
+            indices = np.array(self.dist.rvs(size=size))
+            return self.data[indices]
 
 
 class SexType(IntEnum):
--- a/src/hivpy/sex_behaviour_data.py
+++ b/src/hivpy/sex_behaviour_data.py
@@ -100,6 +100,11 @@ class SexualBehaviourData:
             self.rred_diagnosis = self._get_discrete_dist("rred_diagnosis")
             self.rred_diagnosis_period = self.data["rred_diagnosis"]["Period"]
 
+            self.yearly_risk_behaviour_change = {"1990s": self._get_discrete_dist(
+                                                            "yearly_risk_behaviour_change_90s"),
+                                                 "2010s": self._get_discrete_dist(
+                                                            "yearly_risk_behaviour_change_10s")}
+
         except KeyError as ke:
             print(ke.args)
             raise DataLoadException
--- a/src/hivpy/sex_behaviour_data.py
+++ b/src/hivpy/sex_behaviour_data.py
@@ -47,8 +47,8 @@ class SexualBehaviourData:
             for key, data in prob_dict.items()
         }
 
-    def _select_matrix(self, matrix_list):
-        return matrix_list[np.random.choice(len(matrix_list))]
+    def _select_matrix(self, matrix_list, probabilities):
+        return matrix_list[np.random.choice(len(matrix_list), p=probabilities)]
 
     def __init__(self, filename):
         with open(filename, 'r') as file:
@@ -67,7 +67,8 @@ class SexualBehaviourData:
                 "short_term_partner_distributions", "Sex_Worker")
 
             self.age_based_risk = np.array(
-                self._select_matrix(self.data["age_based_risk_options"]["risk_factor"]))
+                self._select_matrix(self.data["age_based_risk_options"]["risk_factor"],
+                                    self.data["age_based_risk_options"]["Probability"]))
 
             self.sex_behaviour_transition_options = self.data["sex_behaviour_transition_options"]
 
@@ -90,6 +91,8 @@ class SexualBehaviourData:
             self.rred_initial = self.data["rred_initial"]
 
             self.new_partner_dist = self._get_discrete_dist("new_partner_factor")
+
+            self.rred_long_term_partnered = self._get_discrete_dist("rred_partnered")
 
             self.p_rred_p_dist = self._get_discrete_dist("population_rred_personal")
 
--- a/src/hivpy/sex_behaviour_data.py
+++ b/src/hivpy/sex_behaviour_data.py
@@ -66,10 +66,6 @@ class SexualBehaviourData:
             self.sexworker_stp_dists = self._get_discrete_dist_list(
                 "short_term_partner_distributions", "Sex_Worker")
 
-            self.age_based_risk = np.array(
-                self._select_matrix(self.data["age_based_risk_options"]["risk_factor"],
-                                    self.data["age_based_risk_options"]["Probability"]))
-
             self.sex_behaviour_transition_options = self.data["sex_behaviour_transition_options"]
 
             self.sex_mixing_matrix_male_options = self.data["sex_age_mixing_matrices"]["Male"]
@@ -88,13 +84,21 @@ class SexualBehaviourData:
                 SexType.Female: self._get_discrete_dist(
                     "initial_sex_behaviour_probabilities", "Female")}
 
+            # risk reduction factors
             self.rred_initial = self.data["rred_initial"]
+
+            self.age_based_risk = np.array(
+                self._select_matrix(self.data["age_based_risk_options"]["risk_factor"],
+                                    self.data["age_based_risk_options"]["Probability"]))
 
             self.new_partner_dist = self._get_discrete_dist("new_partner_factor")
 
             self.rred_long_term_partnered = self._get_discrete_dist("rred_partnered")
 
             self.p_rred_p_dist = self._get_discrete_dist("population_rred_personal")
+
+            self.rred_diagnosis = self._get_discrete_dist("rred_diagnosis")
+            self.rred_diagnosis_period = self.data["rred_diagnosis"]["Period"]
 
         except KeyError as ke:
             print(ke.args)
--- a/src/hivpy/sex_behaviour_data.py
+++ b/src/hivpy/sex_behaviour_data.py
@@ -105,6 +105,10 @@ class SexualBehaviourData:
                                                  "2010s": self._get_discrete_dist(
                                                             "yearly_risk_behaviour_change_10s")}
 
+            self.rred_art_adherence = self.data["rred_art_adherence"]["Value"]
+            self.adherence_threshold = self.data["rred_art_adherence"]["Adherence_Threshold"]
+            self.rred_art_adherence_probability = self.data["rred_art_adherence"]["Probability"]
+
         except KeyError as ke:
             print(ke.args)
             raise DataLoadException
--- a/src/hivpy/sex_behaviour_data.py
+++ b/src/hivpy/sex_behaviour_data.py
@@ -100,10 +100,8 @@ class SexualBehaviourData:
             self.rred_diagnosis = self._get_discrete_dist("rred_diagnosis")
             self.rred_diagnosis_period = self.data["rred_diagnosis"]["Period"]
 
-            self.yearly_risk_behaviour_change = {"1990s": self._get_discrete_dist(
-                                                            "yearly_risk_behaviour_change_90s"),
-                                                 "2010s": self._get_discrete_dist(
-                                                            "yearly_risk_behaviour_change_10s")}
+            self.yearly_risk_change = {"1990s": self._get_discrete_dist("yearly_risk_change_90s"),
+                                       "2010s": self._get_discrete_dist("yearly_risk_change_10s")}
 
             self.rred_art_adherence = self.data["rred_art_adherence"]["Value"]
             self.adherence_threshold = self.data["rred_art_adherence"]["Adherence_Threshold"]
--- a/src/hivpy/sex_behaviour_data.py
+++ b/src/hivpy/sex_behaviour_data.py
@@ -85,8 +85,8 @@ class SexualBehaviourData:
             self.rred_initial = self.data["rred_initial"]
 
             self.age_based_risk = np.array(
-                self._select_matrix(self.data["age_based_risk_options"]["risk_factor"],
-                                    self.data["age_based_risk_options"]["Probability"]))
+                rng.choice(self.data["age_based_risk_options"]["risk_factor"],
+                           p=self.data["age_based_risk_options"]["Probability"]))
 
             self.new_partner_dist = self._get_discrete_dist("new_partner_factor")
 
--- a/src/hivpy/sexual_behaviour.py
+++ b/src/hivpy/sexual_behaviour.py
@@ -50,6 +50,13 @@ class SexualBehaviourModule:
         }
         self.short_term_partners = {SexType.Male: self.sb_data.male_stp_dists,
                                     SexType.Female: self.sb_data.female_stp_u25_dists}
+        self.ltp_risk_factor = self.sb_data.rred_long_term_partnered
+
+    def update_sex_behaviour(self, population_data):
+        self.num_short_term_partners(population_data)
+        self.update_sex_groups(population_data)
+        self.update_rred_adc(population_data)
+        self.update_rred_balance(population_data)
 
     # Haven't been able to locate the probabilities for this yet
     # Doing them uniform for now
@@ -67,6 +74,7 @@ class SexualBehaviourModule:
         population["rred_age"] = np.ones(n_pop)  # Placeholder to be changed each time step
         population["rred"] = (population["new_partner_factor"] *
                               population["rred_personal"])  # needs * rred_age at each step
+        self.init_rred_adc(population)
 
     def calc_rred_age(self, population, index):
         age = population.loc[index, "age"]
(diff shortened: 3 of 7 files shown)


src/hivpy/common.py
```diff
@@ -19,8 +19,12 @@ def __init__(self, vals, probs):
...
 
     def sample(self, size=1):
-        indices = self.dist.rvs(size=size)
-        return self.data[indices]
+        if(size == 1):
+            index = self.dist.rvs()
+            return self.data[index]
+        else:
+            indices = self.dist.rvs(size=size)
+            return self.data[indices]
```
ageorgou: I think it would be useful to record in the docstring that this returns a single number if size is 1, otherwise an array.

src/hivpy/sexual_behaviour.py
```diff
@@ -47,6 +48,45 @@ def __init__(self, **kwargs):
...
+    def age_index(self, age):
+        return np.minimum((age.astype(int)-self.risk_min_age) //
+                          self.risk_age_grouping, self.risk_categories)
+
+    def update_sex_behaviour(self, population):
+        self.num_short_term_partners(population.data)
+        self.update_sex_groups(population.data)
+        self.update_rred(population)
+
+    def update_rred(self, population):
```
ageorgou: If we're only ever using the population data in these two functions, should we pass in that directly? It may simplify testing too.

src/hivpy/sex_behaviour_data.py
```diff
@@ -84,12 +81,29 @@ def __init__(self, filename):
...
 
+            self.rred_long_term_partnered = self._get_discrete_dist("rred_partnered")
+
             self.p_rred_p_dist = self._get_discrete_dist("population_rred_personal")
 
+            self.rred_diagnosis = self._get_discrete_dist("rred_diagnosis")
+            self.rred_diagnosis_period = self.data["rred_diagnosis"]["Period"]
+
+            self.yearly_risk_change = {"1990s": self._get_discrete_dist("yearly_risk_change_90s"),
+                                       "2010s": self._get_discrete_dist("yearly_risk_change_10s")}
```
ageorgou, line 101: Could we make the keys numerical instead of strings? Perhaps a starting year or a tuple (1990, 1999). It's clear as it is, I just tend to favour keys that can be processed automatically :)

src/hivpy/sexual_behaviour.py
```diff
@@ -47,6 +48,45 @@ def __init__(self, **kwargs):
...
         self.short_term_partners = {SexType.Male: self.sb_data.male_stp_dists,
                                     SexType.Female: self.sb_data.female_stp_u25_dists}
+        self.ltp_risk_factor = self.sb_data.rred_long_term_partnered.sample()
+        self.rred_diagnosis = self.sb_data.rred_diagnosis.sample()
+        self.rred_diagnosis_period = datetime.timedelta(days=365)*self.sb_data.rred_diagnosis_period
+        self.yearly_risk_change = {"1990s": self.sb_data.yearly_risk_change["1990s"].sample(),
+                                   "2010s": self.sb_data.yearly_risk_change["2010s"].sample()}
+        # only need to calculate rred_art_adherence in fraction of runs
+        self.rred_art_adherence_flag = (
+            rng.random() < self.sb_data.rred_art_adherence_probability)
```
ageorgou: I don't fully understand this: do we only update `rred_art_adherence` in some simulations? As it is, it looks like it will either happen every time step or not at all - is that the intent?

ageorgou: Okay, I read the document and this does seem to be the intent! Perhaps something like `rred_consider_art_adherence` would be a more indicative name? Or maybe just expanding on the comment. I just find it a bit counterintuitive, so worth explaining a bit more here. But maybe that's just me!

src/hivpy/sexual_behaviour.py
```diff
@@ -130,3 +144,143 @@ def update_sex_groups(self, population: pd.DataFrame):
...
+        self.init_rred_diagnosis(pop_data)
+        self.init_rred_population(pop_data)
+        self.init_rred_art_adherence(pop_data)
+        self.init_rred_balance(pop_data)
+
+    def init_rred_art_adherence(self, pop_data):
+        pop_data["rred_art_adherence"] = 1
+
+    def update_rred_art_adherence(self, pop_data):
+        if("art_adherence" in pop_data.columns):
```
ageorgou, line 177: Is there any scenario that the column doesn't exist? (same for `partnered` below) Or is this because we have not implemented it yet? (in which case let's mark with a FIXME or note it elsewhere; or just add a column with trivial values!)

src/hivpy/sexual_behaviour.py
```diff
@@ -130,3 +144,143 @@ def update_sex_groups(self, population: pd.DataFrame):
...
                         population.loc[index & jump_to_new_group, "sex_behaviour"] = new_group
+
+    # risk reduction factors
+    def init_risk_factors(self, pop_data):
+        n_pop = len(pop_data)
+        self.init_rred_personal(pop_data, n_pop)
+        pop_data["rred_age"] = np.ones(n_pop)  # Placeholder to be changed each time step
+        self.update_rred_age(pop_data)
+        pop_data["rred"] = (self.new_partner_factor *
+                            pop_data["rred_personal"])
```
ageorgou: Just double checking: the initial `rred` only uses these two factors?
If so, can we move the call to `self.update_rred_age` after the initial calculation, for clarity?

src/hivpy/sexual_behaviour.py
```diff
@@ -130,3 +144,143 @@ def update_sex_groups(self, population: pd.DataFrame):
...
+
+    def init_rred_population(self, pop_data):
+        """Initialise general population risk reduction w.r.t. condomless sex with new partners"""
+        self.rred_population = 1
+
+    def update_rred_population(self, pop_data, date):
+        date1995 = datetime.date(1995, 1, 1)
+        date2000 = datetime.date(2000, 1, 1)
+        date2010 = datetime.date(2010, 1, 1)
+        date2021 = datetime.date(2021, 1, 1)
```
ageorgou: These could be constants on the (Python) module or the class, to avoid recreating them each time.

src/hivpy/sexual_behaviour.py
```diff
@@ -130,3 +144,143 @@ def update_sex_groups(self, population: pd.DataFrame):
...
+        # It prevents needless assignments but requires checking more conditions
+        # Not sure which is more efficient or if it matters.
+        indices = selector(population, rred_adc=(operator.eq, 1), HIV_status=(operator.eq, True))
+        population.loc[indices, "rred_adc"] = 0.2
+
+    def init_rred_population(self, pop_data):
+        """Initialise general population risk reduction w.r.t. condomless sex with new partners"""
+        self.rred_population = 1
+
+    def update_rred_population(self, pop_data, date):
```
ageorgou: Unless I'm missing something!
```suggestion
    def update_rred_population(self, date):
```

src/hivpy/sexual_behaviour.py
```diff
@@ -130,3 +144,143 @@ def update_sex_groups(self, population: pd.DataFrame):
...
+
+    def init_rred_diagnosis(self, population):
+        population["rred_diagnosis"] = 1
+
+    def update_rred_diagnosis(self, population, date):
+        HIV_idx_new = selector(population, HIV_status=(operator.eq, True),
+                               HIV_Diagnosis_Date=(operator.ge, date-self.rred_diagnosis_period))
+        population.loc[HIV_idx_new, "rred_diagnosis"] = self.rred_diagnosis
+        HIV_idx_old = selector(population, HIV_status=(operator.eq, True),
+                               HIV_Diagnosis_Date=(operator.lt, date-self.rred_diagnosis_period))
```
ageorgou, line 243: Just thinking here. I wonder if it's worth recording these variables ("recent diagnosis", "old diagnosis") on the data frame too. But then we'd need to keep the values updated. It may speed this up a bit but probably not significantly. Likely not worth it if they're not used elsewhere (with the same period).

src/hivpy/sexual_behaviour.py
```diff
@@ -130,3 +144,143 @@ def update_sex_groups(self, population: pd.DataFrame):
...
+
+    def init_rred_adc(self, population):
+        population["rred_adc"] = 1.0
+        self.update_rred_adc(population)
+
+    def update_rred_adc(self, population):
+        """Updates risk reduction for AIDS defining condition"""
+        # We don't need the rred_adc==1 condition (since they are all set to rred_adc = 1 to start)
+        # It prevents needless assignments but requires checking more conditions
+        # Not sure which is more efficient or if it matters.
```
ageorgou, line 212: I suspect that avoiding the extra lookup/check will be faster, but not sure at all. I guess the relative efficiency would also depend on how frequent the ADCs are. I'd err in favour of removing the extra check, but may be useful to profile now that the model is becoming more complete.

src/hivpy/sexual_behaviour.py
```diff
@@ -130,3 +144,143 @@ def update_sex_groups(self, population: pd.DataFrame):
...
+        elif partner_discrepancy >= -0.004:
+            rred_balance = 1/0.8
+        elif partner_discrepancy >= -0.005:
+            rred_balance = 1/0.75
+        elif partner_discrepancy >= -0.03:
+            rred_balance = 1/0.7
+        elif partner_discrepancy >= -0.1:
+            rred_balance = 1/0.7
+        else:
+            rred_balance = 10
```
ageorgou: Can we encode the numbers only once, for positive discrepancies, and then check which way we have to adjust? Perhaps in a separate function that gives the ratio when passed a positive discrepancy. So we either use `self.balance_factor(discrepancy)` or `1 / self.balance_factor(-discrepancy)` (and the reciprocal for the other sex)

Looking at the numbers again, maybe it's more complicated than `-discrepancy`, but hopefully the idea is clear!

ageorgou (commented):
I haven't looked through the whole thing yet but had enough comments - will finish soon.

Looks good so far, have left some thoughts.

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
             Expectation = float(n_sex)*p
             N_g = sum(index_sex & index_group)
             assert Expectation - sigma*3 <= N_g <= Expectation + sigma*3
+
+
+def test_rred_long_term_partner():
+    N = 1000
+    pop = Population(size=N, start_date=date(1989, 1, 1))
+    # pick some indices and give those people LTPs
+    indices = [randint(0, N-1) for i in range(15)]
```
ageorgou: We could use the numpy generator for more consistency:
```suggestion
    indices = rng.integers(0, N, size=15)
```
And similar in some other test functions. Confusingly, the numpy functions are exclusive of the high argument, unlike the standard library's.

(it wouldn't really change anything functionally, but as a general guide I suggest we avoid the standard library `random`, so we don't get confused about which seed has been set)

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    pop.data["partnered"] = False
+    pop.data.loc[indices, "partnered"] = True
+    SBM = SexualBehaviourModule()
+    SBM.update_rred_long_term_partnered(pop.data)
+    for i in indices:
+        assert pop.data.loc[i, "rred_long_term_partnered"] == SBM.ltp_risk_factor
```
ageorgou: Could also compare in one go:
```suggestion
    assert all(pop.data.loc[i, "rred_long_term_partnered"] == SBM.ltp_risk_factor)
```

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+
+
+def test_rred_adc():
```
ageorgou, line 122: I like this, think it's a good way of testing the risk factor!

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    SBM.rred_diagnosis_period = timedelta(700)
+    SBM.update_rred_diagnosis(pop.data, pop.date)
+    assert np.all(pop.data["rred_diagnosis"] == 1)
+    # give up some people HIV and advance the date
+    HIV_idx = [randint(0, N-1) for i in range(5)]
+    pop.data.loc[HIV_idx, "HIV_status"] = True
+    pop.data.loc[HIV_idx, "HIV_Diagnosis_Date"] = pop.date
+    SBM.update_rred_diagnosis(pop.data, pop.date)
+    for i in HIV_idx:
+        assert pop.data.loc[i, "rred_diagnosis"] == 4
```
ageorgou: Could also have a "vectorised" assertion here, and later.

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    pop.date += timedelta(days=365)
+    SBM.update_rred_diagnosis(pop.data, pop.date)
+    for i in HIV_idx:
+        assert pop.data.loc[i, "rred_diagnosis"] == 4
+    pop.date += timedelta(days=500)
+    SBM.update_rred_diagnosis(pop.data, pop.date)
+    for i in HIV_idx:
+        assert pop.data.loc[i, "rred_diagnosis"] == 2
```
ageorgou: Can also check that people outside `HIV_idx` still have the old value? (either at the end or after each step)

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    pop = Population(size=N, start_date=date(1989, 1, 1))
+    SBM = SexualBehaviourModule()
+    # need to set the number of new partners to test for risk reduction cases
+    pop.data["num_partners"] = 0
+    men = selector(pop.data, sex=(operator.eq, SexType.Male))
+    women = selector(pop.data, sex=(operator.eq, SexType.Female))
+
+    # equal numbers of partners
+    n_partners_men = n_partners//2
+    n_partners_women = n_partners//2
```
ageorgou: Very minor saving, if you think it reads as clearly:
```suggestion
    n_partners_men = n_partners_women = n_partners//2
```

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    # distribute partners amongst the men and women
+    men_idx = pop.data.loc[men].index
+    women_idx = pop.data.loc[women].index
+    for i in range(n_partners_men):
+        pop.data.loc[rng.choice(men_idx), "num_partners"] += 1
```
ageorgou: Could also do one call to the rng. Not sure if it saves much time though. Only mention it because it seems a bit more natural to me, but not particularly tied to it!
```suggestion
    for i in rng.choice(men_idx, size=n_partners_men):
        pop.data.loc[i, "num_partners"] += 1
```
(it samples with replacement by default, so this should be the same?)

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    SBM.update_rred_diagnosis(pop.data, pop.date)
+    for i in HIV_idx:
+        assert pop.data.loc[i, "rred_diagnosis"] == 4
+    pop.date += timedelta(days=500)
+    SBM.update_rred_diagnosis(pop.data, pop.date)
+    for i in HIV_idx:
+        assert pop.data.loc[i, "rred_diagnosis"] == 2
+
+
+def test_rred_balance():
```
ageorgou, line 169: This may be better as separate tests? Not sure it can be parametrised easily though.

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    SBM.init_risk_factors(pop.data)
+    # art adherence flag off
+    SBM.rred_art_adherence_flag = False
+    pop.data["art_adherence"] = rng.random(size=N)
+    above_idx = selector(pop.data, art_adherence=(operator.ge, 0.8))
+    below_idx = selector(pop.data, art_adherence=(operator.lt, 0.8))
+    SBM.update_rred(pop)
+    for i in range(N):
+        if (pop.data["rred_art_adherence"][i] != 1):
+            print(i, pop.data["rred_art_adherence"][i])
```
ageorgou: Remove if left from debugging?

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    SBM.rred_art_adherence_flag = True
+    SBM.update_rred(pop)
+    assert np.all(pop.data.loc[above_idx, "rred_art_adherence"] == 1)
+    assert np.all(pop.data.loc[below_idx, "rred_art_adherence"] == 2)
+
+    count = 0
+    for i in range(100):
+        SBM_temp = SexualBehaviourModule()
+        count += SBM_temp.rred_art_adherence_flag
+    assert 0.1 < count/100 < 0.3
```
ageorgou: I feel this should be its own test as it's checking something very different from the parts above.

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+
+
+def test_rred_population():
```
ageorgou, line 247: This would also be more concise as a parametrised test, I think.

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    count03, count05, count07 = (0, 0, 0)
+    for i in range(100):
+        SBM = SexualBehaviourModule()
+        SBM.init_risk_factors(pop.data)
+        risk_count = sum(selector(pop.data, rred_personal=(operator.lt, 1)))
+        count03 += (0.2 < (risk_count/N) < 0.4)  # check consistency with threshold
+        count05 += (0.4 < (risk_count/N) < 0.6)  # check consistency with threshold
+        count07 += (0.6 < (risk_count/N) < 0.8)  # check consistency with threshold
+        checks = np.array([x == 1 or x == 1e-5 for x in pop.data["rred_personal"]])
+        assert np.all(checks)  # Check that we only get valid rred values
```
ageorgou: We don't need numpy for this :)
```suggestion
        assert all(x == 1 or x == 1e-5 for x in pop.data["rred_personal"])
```

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    assert(1/6 < count03/100 < 1/2)  # check frequency of threshold from initialisations
+    assert(1/6 < count05/100 < 1/2)  # check frequency of threshold from initialisations
+    assert(1/6 < count07/100 < 1/2)  # check frequency of threshold from initialisations
```
ageorgou, line 300: This feels like quite a broad interval - though I know that's an open discussion!

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    print(risk_factors)
+    assert risk_factors[0] == 1  # no update for under 15
+    assert risk_factors[11] == 1
+    expected_risk_male = pop.sexual_behaviour.age_based_risk.transpose()[0]
+    expected_risk_female = pop.sexual_behaviour.age_based_risk.transpose()[1]
+    assert np.allclose(risk_factors[1:11], expected_risk_male)
+    assert np.allclose(risk_factors[12:22], expected_risk_female)
+    dt = timedelta(days=90)
+    for i in range(20):
+        pop.evolve(dt)
```
ageorgou, line 319: I like how you have set this up to eventually cover all the ages groups.

src/tests/test_sexual_behaviour.py
```diff
@@ -104,3 +106,225 @@ def test_initial_sex_behaviour_groups():
...
+    pop.data["age"] = ages
+    pop.data["sex"] = np.array([SexType.Male]*N + [SexType.Female]*N)
+    pop.sexual_behaviour.init_risk_factors(pop.data)
+    # select a particular risk matrix
+    risk_factors = np.array(pop.data["rred_age"])
+    print(risk_factors)
+    assert risk_factors[0] == 1  # no update for under 15
+    assert risk_factors[11] == 1
+    expected_risk_male = pop.sexual_behaviour.age_based_risk.transpose()[0]
+    expected_risk_female = pop.sexual_behaviour.age_based_risk.transpose()[1]
```
ageorgou: A bit shorter (I think this will be a 2D array, right?):
```suggestion
    expected_risk_male = pop.sexual_behaviour.age_based_risk[:, 0]
    expected_risk_female = pop.sexual_behaviour.age_based_risk[:, 1]
```
or (arrays are unpacked by row):
```suggestion
    expected_risk_male, expected_risk_female = pop.sexual_behaviour.age_based_risk.T
```

src/hivpy/sexual_behaviour.py
mmcleod89: update_rred uses population data and date, so it seemed simpler to me to pass in the whole population object so that there wouldn't be a chance of them being out of sync (e.g. using a date that wasn't associated with that population). 

I'm not sure if passing around population / population data like this is the best approach anyway though, since I think a sexual behaviour module is only going to be associated with one population, we simplify this by just having a member variable point to that population. But then I think the population object has a sexual behaviour module member variable so it starts feeling a bit circular.

src/hivpy/sexual_behaviour.py
mmcleod89: As far as I know this would exist, but we haven't implemented anything around this area yet so I can't guarantee either way. I didn't want to add a trivial column because I didn't want it to be forgotten about later and then have this column being initialised in two places, but I will add a FIXME here that this may be removed once art_adherence has been properly implemented.

src/hivpy/sexual_behaviour.py
mmcleod89: This does actually require an age factor, which wasn't there originally because it was explicit in `update_sex_groups`; I will put the age factor back into the initial rred. 

I'm not really sure why rred is initialised differently (using only three factors) from the way it is updated (using far more); I think the other factors are all initialised to one anyway, but I am not sure if this is because that makes sense from a modelling point of view or because it made sense from a SAS point of view... After all if some people start the simulation with an HIV diagnosis then some of these factors ought to not be one at the start. It might be something worth asking Andrew et al. about in the next call.

src/hivpy/sexual_behaviour.py
mmcleod89: Correct!

src/hivpy/sexual_behaviour.py
mmcleod89: I'm don't know whether that it would speed it up if the categories couldn't be used elsewhere; on the one hand you would only have to update people with a "recent diagnosis" because old diagnoses will stay old; then in this function we only need to update the `rred_diagnosis` value for people who have changed from recent to old in this time step, or from no diagnosis to recent. So to some extent this would then depend on how efficiently we compare the population's current state to previous state. I would be tempted to keep a note of this but stick a pin in it for now?

src/tests/test_sexual_behaviour.py
mmcleod89: I assume here `i` --> `indices` ?

src/tests/test_sexual_behaviour.py
mmcleod89: Yes, that's partly because in order to keep all of these tests from taking a very long time I was tending towards a small population size. Expectation value = 33.3 with variance 22.2... Naturally we can increase the population size and decrease the interval for a more precise test but we might end up in a situation where we have a lot of tests creating and evolving large populations and things taking a long time.

src/hivpy/sex_behaviour_data.py
mmcleod89: I'm happy to make these numerical if you like, I'm just not sure what you mean by keys that can be processed automatically? I wouldn't have thought there was much difference in the type of key as long as it's hashable?

src/hivpy/common.py
```diff
@@ -18,13 +18,16 @@ def __init__(self, vals, probs):
         self.data = vals
         self.dist = stat.rv_discrete(values=(index_range, probs))
 
-    def sample(self, size=1):
-        if(size == 1):
+    def sample(self, **kwargs):
```
ageorgou: To avoid allowing the user to pass in arbitrary arguments (which will be unused), it may be better to define a single optional argument (`size=None`) and then check `if size is not None`. That's also more similar to how scipy/numpy do it, I think?

src/hivpy/sexual_behaviour.py
ageorgou: Oh right, I missed that the date was used.

I've also been wondering if the circular references are strange. We can consider that in the future.

src/hivpy/sex_behaviour_data.py
ageorgou: I meant in case we need to find which decade we are in, for instance, and match to the relevant yearly risk. Having the keys as numbers would make it easier to retrieve the relevant entry. But that's probably more trouble than it's worth.

ageorgou (approved):
Looks good, just one suggestion for sampling from the discrete variable class.

I think the outstanding questions are:
- Factors that go into the initialisation of `rred` variables (vs the update) #47
- Ensuring that we create `art_adherence` #48
- "Meta-variables" like "newly_diagnosed": which ones are worth having? #49

I'll open some issues to track those separately.

src/hivpy/common.py
mmcleod89: The scipy implementation uses an arbitrary **kwargs pattern, but I agree that your suggestion is better as it tries to enforce some correctness and at least will make it clear to them if someone throws the wrong arguments at it.

Issue #31, opened 2022-02-03
Review and Config of Simulation Output


pweschmidt:
Should it be removed from the SimulationConfig?

- which place in the Config file
- how should it be passed into the code 
- eliminate the track attributes

NovusEdge:
Hi There!

I would like to take this issue if no one's assigned
Please let me know the specifics (if any).

ageorgou:
- Remove tracked attributes completely in favour of `SimulationOutput`
- Restructure config file:
    - Switch to YAML
    - Move output directory setting out of simulation config; create an output section (with subsections for logging, simulation, ...?)
    - Pass simulation output path to `SimulationHandler` constructor
- Remove `run_simulation` (replace with calls to `SimulationHandler`)
- `SimulationOutput`: remove output path, only pass in as argument to `write_output`

PR #50, opened 2022-05-24
Enhancement/output config

mmcleod89:


src/hivpy/cli.py
```diff
@@ -18,21 +19,24 @@
...
+    parser.add_argument("input", type=pathlib.Path, help="run_model config.yaml")
     args = parser.parse_args()
-    conf_filename = args.input
-    config = configparser.ConfigParser()
+    config_filename = args.input
+    try:
+        with open(config_filename, 'r') as file:
+            config = yaml.safe_load(file)
+    except yaml.YAMLError as err:
+        print("Error parsing yaml file {}".format(err))
```
ageorgou, line 33: Maybe we can print the error in a more readable way, but probably fine to leave this until we get actual errors and feedback from users.

src/hivpy/simulation.py
```diff
@@ -82,38 +79,28 @@ def update_summary_stats(self, date, pop_data):
...
     population: Population
-    results: pd.DataFrame
 
     def __init__(self, simulation_config):
         self.simulation_config = simulation_config
-        self.results = None
         self._initialise_population()
         self.output = SimulationOutput(self.simulation_config)
+        self.output_path = simulation_config.output_dir / (
+            "simulation_output." + str(datetime.now().strftime("%Y%m%d-%H%M%S")) + ".csv")
```
ageorgou: Unrelated to these changes, but it might be better to have the filename as `simulation_output_{date}.csv` instead, to avoid the extra `.` which can be confusing/mess up how the name is displayed?

src/tests/test_experiment.py
```diff
@@ -1,19 +1,19 @@
...
     """Sample experiment parameters for testing."""
-    parser = ConfigParser()
-    parser.read(
-        os.path.join(os.path.dirname(__file__), 'fixtures', 'sample.conf'))
-    print(list(parser.items()))
-    return parser
+    filepath = os.path.join(os.path.dirname(__file__), 'fixtures', 'sample.yaml')
+    with open(filepath, 'r') as sample_file:
+        sample_config = yaml.safe_load(sample_file)
+    print(sample_config)
```
ageorgou: Oops, this was probably left in from debugging the config parser in the past!

src/tests/test_simulation.py
```diff
@@ -43,40 +44,21 @@ def test_error_end_before_first_step(tmp_path):
...
-    pop, results = run_simulation(config)
+    config = SimulationConfig(size, start, end, tmp_path, step)
+    simulation_handler = SimulationHandler(config)
+    simulation_handler.run()
+    pop = simulation_handler.population
     # Check that the number alive never grows... (some steps may have 0 deaths)
-    assert all(results.num_alive.diff()[1:] <= 0)
+    # assert all(results.num_alive.diff()[1:] <= 0)
+    num_not_dead = sum(pop.data.date_of_death.isnull())
+    assert num_not_dead < len(pop.data)
```
ageorgou, line 61: Can we check this from the SimulationOutput instead? (If we don't store the number of people alive there, we should! though that can be a different PR)

hivpy.yaml
```diff
@@ -0,0 +1,12 @@
+EXPERIMENT: 
+  population: 1000
+  start_year: 1989
+  end_year: 2021
+  time_interval_days: 90
+  simulation_output_dir: output
```
ageorgou, line 6: Looking back at [the issue comment](https://github.com/UCL/hivpy/issues/31#issuecomment-1131793090), I'd noted that we could move this out of this section and instead have a general "Output" section with the logging and other output options. I don't see any problem doing it this way either (probably better for consistency too), just writing it down so there's a record if we look back in the future!

ageorgou (approved):
All looks good, just some small comments. We should also delete the `hivpy.example.conf` file at the top level.

src/hivpy/simulation.py
mmcleod89: It shouldn't cause any issues for the name being displayed (it doesn't on my Windows machine at least) but I don't mind changing the style (in which case we should change the style of the log output files as well which are done similarly).

src/tests/test_experiment.py
mmcleod89: Okay I will remove it!

src/tests/test_simulation.py
mmcleod89: I think we can add in something that checks the simulation output too, but I'm inclined to also check that the population object is itself updating correctly as that's the thing that's going to be accessed by other parts of the simulation that are affected by this.

Issue #38, opened 2022-04-07
Clarify how data are passed between different parts of the model


pweschmidt:
* Parameters loaded from data files
* how modules are passing data between them (e.g. via data frames?)

ageorgou:
Related: should we allow setting parameters for the modules directly through the constructor, or only by modifying an associated object (e.g. `DemographicsData`)?

PR #52, opened 2022-05-25
Read some demographics from YAML

### LINKED ISSUE #44: Static files for death rates
Following from #17, we should read in the rates from a file rather than hardcode them.

ageorgou:
Fixes #44.

- Put death rates and sex ration in YAML file
- Extend tests for balance of sexes

Have tested the "new" values against the old code with the same seed and the death results for a single step are the same. Haven't included the test as it's hard to make sense of the results.

The `DemographicsData` is quite small at the moment, but I think it was worth having as there will likely be more parameters to come, and for the abstraction. Not all parameters are moved there yet, and the old `params` dictionary still exists, for ease of overriding; I'm not sure if we should allow overriding them directly on the `DemographicsModule` constructor or only by passing in a `DemographicsData` object (see also comment in #38).

There's also maybe a discussion to be had about a parent class for Modules and ModuleData, if we want to enforce consistency or some base behaviour across modules.

### DIFF
--- a/src/hivpy/demographics_data.py
+++ b/src/hivpy/demographics_data.py
@@ -1,0 +1,19 @@
+import yaml
+
+from .common import SexType
+
+
+class DemographicsData:
+    """A class for holding demographics-related data loaded from a file."""
+
+    def __init__(self, filename):
+        with open(filename, 'r') as file:
+            data = yaml.safe_load(file)
+
+        self.female_ratio = data["female_ratio"]
+
+        self.death_age_limits = data["death_rates"]["Age_limits"]
+        self.death_rates = {
+            SexType.Female: [0] + data["death_rates"]["Female"],
+            SexType.Male: [0] + data["death_rates"]["Male"]
+        }
--- a/data/demographics.yaml
+++ b/data/demographics.yaml
@@ -1,0 +1,8 @@
+female_ratio: 0.52
+
+death_rates:
+  # The lower end of each age group: (ages below the first group assumed to have 0 death rate)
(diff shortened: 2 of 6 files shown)


src/hivpy/demographics_data.py
```diff
@@ -0,0 +1,19 @@
...
+class DemographicsData:
+    """A class for holding demographics-related data loaded from a file."""
+
+    def __init__(self, filename):
+        with open(filename, 'r') as file:
+            data = yaml.safe_load(file)
+
+        self.female_ratio = data["female_ratio"]
+
+        self.death_age_limits = data["death_rates"]["Age_limits"]
```
mmcleod89, line 15: Does this need a `+ [inf]` as in the previous `demographics.py:258`?

mmcleod89 (commented):
I think that we need some kind of a test to check that roughly the right proportion of people are dying in different age groups in a particular time step, e.g. set up a population which has 1000 people in each age group and check how many of each age group die in one time step. 

I notice that the `determine_deaths` method hasn't really changed, but the age limits and the death rates have (age limits doesn't have an inf at the end; death rates now has a 0 at the front). It looks correct to me, but was it not working before? It seems like this should change the result. 

I agree that it's good to add the demographic_data module now even though it's small; we should also move the age demographic data into it, but that might be another pull request?

I also wonder if in line 203 of `demographics.py` we could change `age_groups` to `demographic_groups` or something similar, since it is both age and sex?

mmcleod89 (approved):

PR #54, opened 2022-05-30
Refer to location of data files more robustly

### LINKED ISSUE #51: Data location not specified correctly
Because of how we give the path to the data files such as `sexual_behaviour_data.yaml`, the code only works if executed from the top-level directory. We should:
- move the data into the package directory
- (if needed) edit the setup configuration to ensure the data files are also installed along with the code
- use [`importlib.resources`](https://docs.python.org/3.8/library/importlib.html?highlight=importlib#module-importlib.resources) to get the path to wherever the files have been installed.

ageorgou:
Fixes #51.

The files are now searched for relative to where the package is installed, not where we run the code. Also some small related changes to the tests.


mmcleod89 (approved):
Looks good; I don't have any issues with this. I think it raises another question though, which is the proper use of data files. I think that the reading of these default data files was originally a place holder, with the end point being that in the configuration you could set where the data files are, so that you can keep track of all of the parameters you used for different experiments without having to change any code. I think the way that this PR works would be a good basis for a set of default data files, and we can (as a separate issue) add functionality to use different data files as well, with the assumption being that the data files in the installation maintain a kind of standard rather than being edited to tinker with parameters?

ageorgou:
Yes, that's a good point. We would want a way (in the config, probably) of specifying e.g. a custom demographics data file (rather than individual parameters, which could get long very quickly). Then instantiate the demographics module using that location rather than the default one. Also related to #38 - do we switch to passing a `DemographicsData` into the `DemographicsModule` constructor or the other way around (pass the path to the data file from the module to the data constructor)?

ageorgou:
To be followed up in #55.

PR #59, opened 2022-06-29
Refactor population column names

### LINKED ISSUE #56: Column names should be constants
To live in their own module?

mmcleod89:
Move population column names from string literals to variables in a separate file.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,4 +47,5 @@ max-line-length = 100
 per-file-ignores =
     # Don't complain about unused imports in __init__.py
     */__init__.py: F401
+    # Ignore max line length as some column names and descriptions are very long
     */column_names.py: E501
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,3 +47,4 @@ max-line-length = 100
 per-file-ignores =
     # Don't complain about unused imports in __init__.py
     */__init__.py: F401
+    */column_names.py: E501
--- a/src/hivpy/column_names.py
+++ b/src/hivpy/column_names.py
@@ -1,0 +1,20 @@
+# Column names in the Population Data-Frame
+
+SEX = "sex"                                 # common.SexType: SexType.Male for male and SexType.Female for female
+AGE = "age"                                 # float: age at the current date
+
+RRED = "rred"                               # float: overall rred value from combined factors
+RRED_AGE = "rred_age"                       # float: risk reduction factor based on age
+RRED_ADC = "rred_adc"                       # float: risk reduction for AIDS defining condition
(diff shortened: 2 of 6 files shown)


mmcleod89:
The way that I've done the comments in the column file currently will fail static checks, but I think looks much neater than putting the comment on the line above each variable declaration. 

Does Python have genuine support for constants?

src/hivpy/demographics.py
```diff
@@ -6,6 +6,7 @@
 import pandas as pd
 from scipy.interpolate import interp1d
 
+import hivpy.column_names as col
```
ageorgou, line 9: I wonder if it's preferable to import specific names directly (`from hivpy.column_names import AGE_GROUP, SEX`), to:
- save some typing
- see at a glance which columns are referenced in a particular file

I don't have strong feelings either way though.

src/hivpy/column_names.py
```diff
@@ -0,0 +1,21 @@
+# Column names in the Population Data-Frame
```
ageorgou: This seems a good example of a module-level docstring.
```suggestion
"""Column names in the Population Data-Frame."""
```

ageorgou (commented):
Looks good. No, there is strangely no direct support for constants, although there are ways around that.

I can think of two ways of making the static checks happy:
- Disabling the line length check for the column names file specifically (though there are also a couple of errors for trailing whitespace); [we do something similar for another file at the moment](https://github.com/UCL/hivpy/blob/da264365f33cdcbcca96b50938fecac805543e7a/setup.cfg#L45-L49)
- Only using two spaces after each column name, which breaks the alignment.

With the second one, there's still risk that some lines will be too long. We could shorten them by removing the type, though that's useful to have. On the other hand, I don't know if the types fit fully with the names or are better placed when we create the DataFrame. On the hypothetical third hand, we don't list all the columns there but create them later, so putting the types there wouldn't work as it is anyway.

Overall, maybe ignoring the line length on this file is the simplest option. It's more of a data file than code, anyway, so we don't have the same readability concerns.

setup.cfg
```diff
@@ -47,3 +47,4 @@ max-line-length = 100
 per-file-ignores =
     # Don't complain about unused imports in __init__.py
     */__init__.py: F401
+    */column_names.py: E501
```
ageorgou, line 51: I'd write out what we ignore here because it's not clear from the error code without looking it up. For example:

```suggestion
    # Ignore max line length as some column names and descriptions are very long
    */column_names.py: E501
```
But not major!

ageorgou (approved):

PR #64, opened 2022-07-18
Remove attributes from population class

### LINKED ISSUE #58: Remove `attributes` from Population class
Population class currently has a dictionary called `attributes` which calculates some summary values at initialisation but isn't used thereafter, which was previously used for outputs. The tracked outputs have been removed, so this is now redundant and should be removed. 

If we want to use similar functionality (for example, having the number of people alive calculated at each time-step to avoid re-calculating it in different parts of the code) then we can re-implement this as a field or method.

mmcleod89:


ageorgou (commented):
Looks good, just need to fix the relevant linting failures:
> ./src/hivpy/population.py:2:1: F401 'functools' imported but unused
> ./src/hivpy/population.py:3:1: F401 'typing.Callable' imported but unused
> ./src/hivpy/population.py:3:1: F401 'typing.Dict' imported but unused
> ./src/hivpy/population.py:9:1: F401 '.common.SexType' imported but unused

(the other errors should be fixed by updating to the current `development`)

ageorgou (approved):

PR #65, opened 2022-07-25
Population transforms

mmcleod89:
Use pandas transforms for modifying populations. (In common.py in this branch.)

### DIFF
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -8,6 +8,8 @@ from functools import reduce
 import numpy as np
 import scipy.stats as stat
 
+import hivpy.column_names as col
+
 
 class DiscreteChoice:
     def __init__(self, vals: np.ndarray, probs):
@@ -43,6 +45,15 @@ def selector(population, **kwargs):
     index = reduce(operator.and_,
                    (op(population[kw], val) for kw, (op, val) in kwargs.items()))
     return index
+
+
+def transform_group(df, param_list, func):
+    """Groups the data by a list of parameters and applies a function to each grouping"""
+    # HIV_STATUS is just a dummy column to allow us to use the transform method
+    def general_func(g):
+        args = [n for n in g.name] + [g.size]
+        return func(args)
+    return df.groupby(param_list)[col.HIV_STATUS].transform(general_func)
 
 
 def between(values, limits):
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -59,7 +59,7 @@ def transform_group(df, param_list, func, use_size=True):
     # HIV_STATUS is just a dummy column to allow us to use the transform method
     def general_func(g):
         args = list(g.name)
-        if(use_size):
+        if (use_size):
             args.append(g.size)
         return func(*args)
     return df.groupby(param_list)[col.HIV_STATUS].transform(general_func)
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -51,7 +51,7 @@ def transform_group(df, param_list, func, use_size=True):
     """Groups the data by a list of parameters and applies a function to each grouping"""
     # HIV_STATUS is just a dummy column to allow us to use the transform method
     def general_func(g):
-        args = [n for n in g.name] # list(g.name)
+        args = list(g.name)
         if(use_size):
             args.append(g.size)
         return func(*args)
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -48,7 +48,14 @@ def selector(population, **kwargs):
 
 
 def transform_group(df, param_list, func, use_size=True):
-    """Groups the data by a list of parameters and applies a function to each grouping"""
+    """Groups the data by a list of parameters and applies a function to each grouping. \n
+    `param_list` is a list of names of columns by which you want to group the data. The order must
+    match the order of arguments taken by the function `func` \n
+    `func` is a function which takes the values of those columns for a group (and optionally the
+    size of the group, which should be the last argument) and returns a value or array of values of
+    the size of the group. \n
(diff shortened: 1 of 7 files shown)


src/hivpy/sexual_behaviour.py
```diff
@@ -109,37 +110,48 @@ def prob_transition(self, sex, rred, i, j):
...
-        for sex in SexType:
-            for g in SexBehaviours[sex]:
-                index = selector(population, sex=(operator.eq, sex),
-                                 sex_behaviour=(operator.eq, g),
-                                 age=(operator.gt, 15))
-                population.loc[index, col.NUM_PARTNERS] = (
-                    self.short_term_partners[sex][g].sample(size=sum(index)))
+        """Calculate the number of short term partners for the whole population"""
+        # can we avoid doing population.loc[active_pop] twice? Does it waste time?
+        active_pop = population.index[(15 <= population.age) & (population.age < 65)]
```
ageorgou, line 122: This may be a useful method on the `Population`, since it's also computed when updating sex groups. Maybe a method that also does the `.loc` and returns the reduced dataframe directly.

src/tests/test_sexual_behaviour.py
```diff
@@ -71,18 +95,26 @@ def check_num_partners(row):
 
 def test_num_partners():
     """Check that number of partners are reasonable"""
-    pop_data = Population(size=1000, start_date=date(1989, 1, 1)).data
-    assert(any(pop_data["num_partners"] > 0))
+    pop = Population(size=100000, start_date=date(1989, 1, 1))
+    t1 = time.perf_counter()
```
ageorgou: These (along with `import time`) can be removed now, or is performance a concern?

src/hivpy/demographics.py
```diff
@@ -195,9 +195,7 @@ def initialise_age(self, count):
 
         return age_distribution.gen_ages(count)
 
-    def _probability_of_death(self, args) -> float:
-        sex = args[0]
-        age_group = args[1]
+    def _probability_of_death(self, sex, age_group) -> float:
```
ageorgou: We've lost the type hints during the change in signature and back!

src/tests/test_sexual_behaviour.py
```diff
@@ -71,18 +93,23 @@ def check_num_partners(row):
 
 def test_num_partners():
     """Check that number of partners are reasonable"""
-    pop_data = Population(size=1000, start_date=date(1989, 1, 1)).data
-    assert(any(pop_data["num_partners"] > 0))
+    pop = Population(size=100000, start_date=date(1989, 1, 1))
+    # t1 = time.perf_counter()
+    # for x in range(0, 500):
```
ageorgou: Remove these (and the call to `import time`) instead of commenting, or is performance still a concern?

src/tests/test_demographics.py
```diff
@@ -91,11 +92,11 @@ def test_death_rate():
     group_size = N // 2 // len(ages_to_try)
     ages = sum(([age] * group_size for age in ages_to_try), []) * 2
     sexes = [SexType.Female] * (N // 2) + [SexType.Male] * (N // 2)
-    population_data = pd.DataFrame({
-        'date_of_death': [None] * N,
-        'age': ages,
-        'sex': sexes
-    })
+    pop = Population(size=N, start_date=date(1989, 1, 1))
```
ageorgou, line 95: We may want to think about having a pytest fixture that creates a population with some characteristics. Not a question for here/now though!

ageorgou (changes requested):
The changes themselves look reasonable but this seems to have some clashes with `development` due to being a few commits behind. Could rebase/merge after merging #64?

src/hivpy/common.py
```diff
@@ -45,6 +47,17 @@ def selector(population, **kwargs):
     return index
 
 
+def transform_group(df, param_list, func, use_size=True):
+    """Groups the data by a list of parameters and applies a function to each grouping"""
```
ageorgou: I'd also suggest expanding this a bit to explain the use_size.

src/hivpy/sexual_behaviour.py
mmcleod89: I think it makes more sense to have this fixed in the refactor pull request, because it works a little differently once the `transform_group` method has been moved into `Population` anyway, and then you do only do the `loc` once.

This raised in my mind the possibility of exchanging `common.selector` with something like this:
```
def get_subpopulation_data(self, conditions):
        """Gets a dataframe for sub-population based on a list of conditions. \n
        Conditions are in the format: (column, operator, value) \n
        e.g. (col.AGE, operator.ge, 15) to enforce age >= 15"""
        index = reduce(operator.and_, (op(self.data[column], val) for (column, op, val) in conditions))
        return index
```
which could live in the `Population` class and allows there to be multiple conditions on a given column (for example `age >= 15`, and `age < 65`) which currently isn't possible when using dictionary keywords which prevents repeated keys.

src/tests/test_sexual_behaviour.py
mmcleod89: I still had these in place at the time in case they needed to be used again during the testing of the refactor. In the refactor they've been commented out but can be removed completely; I wasn't sure if we might want to be able to quickly uncomment them and check the timing if we continued to change how the the dataframe is updated in the future and wanted to be able to check the performance impact on different parts of the code, but perhaps this can be done better by creating some kind of timing function that can be inserted into tests easily..

ageorgou: Yeah, as discussed, I'd rather leave this a profiler or similar rather than have the timing in the code. If you think they'll be needed immediately, can leave them in and remove them in the refactor in #66.

ageorgou (approved):

PR #66, opened 2022-07-25
Refactor transforms

mmcleod89:
Move transforms into the population class. (Builds on some work from population-transforms branch.) 

Had to add **kwards to transform_group in order to accommodate acting on sub-populations, although there may be a better way to do this. 

The code can be a little confusing the way it's currently done because Population class can't be imported into SexualBehaviour or Demographics modules due to circular reference, which isn't ideal. Could be sorted out by moving those modules into the simulation and letting them operate on the population class instead of being part of the population class.

### DIFF
--- a/src/hivpy/population.py
+++ b/src/hivpy/population.py
@@ -87,6 +87,16 @@ class Population:
         """Get the value of the named attribute at the current date."""
         return self.attributes[attribute_name](self.data)
 
+    def transform_group(self, param_list, func, use_size=True):
+        """Groups the data by a list of parameters and applies a function to each grouping"""
+        # HIV_STATUS is just a dummy column to allow us to use the transform method
+        def general_func(g):
+            args = [n for n in g.name] # list(g.name)
+            if(use_size):
+                args.append(g.size)
+            return func(*args)
+        return self.data.groupby(param_list)[col.HIV_STATUS].transform(general_func)
+
     def evolve(self, time_step: datetime.timedelta):
         """Advance the population by one time step."""
         # Does nothing just yet except advance the current date, track ages
--- a/src/hivpy/population.py
+++ b/src/hivpy/population.py
@@ -1,11 +1,9 @@
 import datetime
-import functools
 
 import pandas as pd
 
 import hivpy.column_names as col
 
-from .common import SexType
 from .demographics import DemographicsModule
 from .hiv_status import HIVStatusModule
 from .sexual_behaviour import SexualBehaviourModule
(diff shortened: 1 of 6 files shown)


src/hivpy/population.py
```diff
@@ -51,15 +51,29 @@ def _create_population_data(self):
         self.data[col.NUM_PARTNERS] = 0
         self.sexual_behaviour.init_sex_behaviour_groups(self.data)
         self.sexual_behaviour.init_risk_factors(self.data)
-        self.sexual_behaviour.num_short_term_partners(self.data)
+        self.sexual_behaviour.num_short_term_partners(self)
+
+    def transform_group(self, param_list, func, use_size=True, **kwargs):
+        """Groups the data by a list of parameters and applies a function to each grouping"""
```
ageorgou: The rest of the docstring seems to have been lost!

src/hivpy/population.py
```diff
@@ -51,15 +51,29 @@ def _create_population_data(self):
...
+        # HIV_STATUS is just a dummy column to allow us to use the transform method
+        def general_func(g):
+            args = list(g.name)
+            if (use_size):
+                args.append(g.size)
+            return func(*args)
+        if "sub_pop" in kwargs.keys():
+            df = self.data.loc[kwargs["sub_pop"]]
```
ageorgou: Since this is a fixed keyword argument that we (potentially) expect, we could make it explicitly part of the signature with a default value, like `use_size`. I think that would make it cleaner for people looking at how to call the function too.

ageorgou (changes requested):
I think this new way of calling looks better, especially when applying to a subpopulation. Only some small comments here, otherwise looks ready.

One thing that may make this simpler is adding some methods to the `Population` for e.g. reading or writing columns. That way we wouldn't need to go through the `.data` each time, and we could change the implementation more easily in the future. But that's for another PR...

src/hivpy/population.py
mmcleod89: I did think about this, but the problem that I have here is that for something like `use_size` there is a trivial default, and I wasn't sure how to define a default so that the default sub-population is the whole population, and then that would lead to an unnecessary call to `loc` anyway. But I suppose what we could do is have a default `sub_pop=None`? And then presumably can test for `None` and apply `loc` only if the `sub_pop` is defined.

ageorgou: Yes, that's what I had in mind. `None` is a common default in many Python frameworks to indicate that extra processing is not needed.

ageorgou (approved):

Issue #62, opened 2022-07-05
HIV transmission


ageorgou:
From both long-term and short-term partners.

- [ ] #154 
- [ ] #90

PR #69, opened 2022-08-18
Introduce HIV into the population

### LINKED ISSUE #63: Initial presence of HIV
Reflect distribution from SAS model.

ageorgou:
Resolves #63.

I'm not sure whether the "starting" date is better placed in the population or the HIV status module (or somewhere else).

I've adjusted the transmission rate to make the test more reliable, but that part will be overwritten anyway (#62).

One thing to check: in the SAS model, this occurs in "year" 1989.25, which would be April 1, not January 1. Does this make a difference in this implementation (if we start a simulation on January 1)? Or are all the variables set up for it?

### DIFF
--- a/src/hivpy/hiv_status.py
+++ b/src/hivpy/hiv_status.py
@@ -1,6 +1,5 @@
 import operator
 
-import numpy as np
 import pandas as pd
 
 import hivpy.column_names as col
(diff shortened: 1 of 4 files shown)


src/hivpy/hiv_status.py
```diff
@@ -10,18 +9,24 @@
...
+        """Initialise HIV status at the start of the simulation to no infections."""
+        # This may be useful as a separate method if we end up representing status
+        # as something more complex than a boolean, e.g. an enum.
+        return pd.Series(False, population.index)
+
+    def introduce_HIV(self, population: pd.DataFrame):
+        """Initialise HIV status at the start of the pandemic."""
+        # At the start of the epidemic, we consider only people with short-term partners over
+        # the threshold as potentially infected.
+        initial_candidates = population[col.NUM_PARTNERS] >= self.initial_hiv_newp_threshold
```
mmcleod89, line 25: Does this work correctly without using `index`? I don't think the tests cover a case where there are people both above and below the threshold number of partners so it doesn't check if the HIV status is being applied to the correct people in a mixed population.

mmcleod89 (commented):
In general I think this is fine, as long as we can be sure that the HIV status is being assigned to the correct people in the population when it is not everyone/no-one. 

The change to the `HIV_infection_risk` will cause a merge conflict I think with the branch that I'm working on, but wouldn't be hard to fix.

src/hivpy/hiv_status.py
ageorgou: Good point, I'd thought about that but not implemented it. I've changed the tests so we also check a mixed population.

mmcleod89 (approved):

ageorgou:
Test failing because HIV transmission no longer corresponds to test, but this will be overwritten in #70.

Issue #48, opened 2022-05-19
Ongoing: undefined population variables


ageorgou:
A list of variables we are referring to but are either not defined or trivially initialised.

Ideally this will not happen, and we will only introduce a variable in the code when we are sure that it can be initialised properly. However, on some occasions, referring to a variable before initialisation (usually with some condition that checks whether it has been defined) may outweigh the risks. In those cases, we should make a note here so that we do not accidentally forget to include it:

- [ ] `art_adherence` (introduced in #46, fixed in ??)

mmcleod89:
- Viral Load Group; currently dummy initialised (uniform random) in HIV status module. Should be implemented when we start to look at HIV progression / treatment.

PR #70, opened 2022-08-19
WIP: HIV transmission for short term partners 

mmcleod89:
Still needs tests to check that things are correct, which might benefit from some refactoring. Not all factors for HIV transmission are accounted for yet, as they don't exist in the model, but I did add a dummy initialisation for viral load groups since that was the most complicated effect and central to the calculation rather than just being a post-factor. 

We will need a data file for transmission properties too but I think this should be a separate PR

### DIFF
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -35,6 +35,8 @@ class SexType(IntEnum):
     Male = 0
     Female = 1
 
+def opposite_sex(sex: SexType):
+    return (1 - sex)
 
 def selector(population, **kwargs):
     """Select the rows of a population data frame matching a set of criteria.
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -35,8 +35,10 @@ class SexType(IntEnum):
     Male = 0
     Female = 1
 
+
 def opposite_sex(sex: SexType):
     return (1 - sex)
+
 
 def selector(population, **kwargs):
     """Select the rows of a population data frame matching a set of criteria.
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -15,6 +15,7 @@ class DiscreteChoice:
         if (len(probs) != N):
             raise Exception
         index_range = np.arange(0, N, 1)
+        self.probs = probs
         self.data = vals
         self.dist = stat.rv_discrete(values=(index_range, probs))
 
--- a/src/hivpy/hiv_status.py
+++ b/src/hivpy/hiv_status.py
@@ -45,18 +45,17 @@ class HIVStatusModule:
         # Should we be using for loops here or can we do better?
         for sex in SexType:
             for age_group in range(5):   # FIXME need to get number of age groups from somewhere
-                sub_pop_indices = population.data.index[(population.data[col.SEX] == sex) & (
+                sub_pop = population.data.loc[(population.data[col.SEX] == sex) & (
                     population.data[col.SEX_MIX_AGE_GROUP] == age_group)]
-                sub_pop = population.data.loc[sub_pop_indices]
                 # total number of people partnered to people in this group
                 n_stp_total = sum(sub_pop[col.NUM_PARTNERS])
-                # num people parters to HIV+ people in this group
-                n_infected = sum(sub_pop.loc[sub_pop[col.HIV_STATUS]][col.NUM_PARTNERS])
+                # num people partered to HIV+ people in this group
+                n_stp_of_infected = sum(sub_pop.loc[sub_pop[col.HIV_STATUS], col.NUM_PARTNERS])
                 # Probability of being HIV prositive
-                if n_infected == 0:
+                if n_stp_of_infected == 0:
                     self.stp_HIV_rate[sex][age_group] = 0
                 else:
-                    self.stp_HIV_rate[sex][age_group] = n_infected / \
+                    self.stp_HIV_rate[sex][age_group] = n_stp_of_infected / \
                         n_stp_total  # TODO: need to double check this definition
                 # Chances of being in a given viral group
                 if n_stp_total > 0:
@@ -101,7 +100,7 @@ class HIVStatusModule:
         viral_transmission_probabilities = np.array([max(0, rng.normal(
             self.transmission_means[group], self.transmission_sigmas[group]))
             for group in stp_viral_groups])
-        if person[col.SEX] == SexType.Female:
+        if person[col.SEX] is SexType.Female:
             if person[col.AGE] < 20:
                 viral_transmission_probabilities = (viral_transmission_probabilities
                                                     * self.fold_change_yw)
--- a/src/hivpy/hiv_status.py
+++ b/src/hivpy/hiv_status.py
@@ -88,13 +88,14 @@ class HIVStatusModule:
                     infection_prob[i] *= self.fold_change_yw
                 else:
                     infection_prob[i] *= self.fold_change_w
-            return infection_prob
+        return infection_prob
 
     def stp_HIV_transmission(self, person):
         # TODO: Add circumcision, STIs etc.
         """Returns True if HIV transmission occurs, and False otherwise"""
-        stp_viral_groups = np.array([rng.choice(7, p=self.stp_viral_group_rate[opposite_sex(
-            person[col.SEX])][age_group]) for age_group in person[col.STP_AGE_GROUPS]])
+        stp_viral_groups = np.array([
+            rng.choice(7, p=self.stp_viral_group_rate[opposite_sex(person[col.SEX])][age_group])
+            for age_group in person[col.STP_AGE_GROUPS]])
         HIV_probabilities = np.array([self.stp_HIV_rate[opposite_sex(
             person[col.SEX])][age_group] for age_group in person[col.STP_AGE_GROUPS]])
         viral_transmission_probabilities = np.array([max(0, rng.normal(
--- a/src/hivpy/hiv_status.py
+++ b/src/hivpy/hiv_status.py
@@ -17,13 +17,13 @@ class HIVStatusModule:
     def __init__(self):
         self.stp_HIV_rate = {SexType.Male: np.zeros(5),
                                  SexType.Female: np.zeros(5)}  # FIXME
-        self.stp_viral_group_rate = {SexType.Male: np.array([np.zeros(6)]*5),
-                                     SexType.Female: np.array([np.zeros(6)]*5)}
+        self.stp_viral_group_rate = {SexType.Male: np.array([np.zeros(7)]*5),
+                                     SexType.Female: np.array([np.zeros(7)]*5)}
         # FIXME move these to data file 
         self.fold_tr_newp = rng.choice([0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1/0.8, 1/0.6, 1/0.4])  # a more descriptive name would be nice
-        self.fold_change_w = rng.choice([1, 1.5, 2], p=[0.05, 0.25, 0.7])
-        self.fold_change_yw = rng.choice([1, 2, 3]) * self.fold_change_w
-        self.fold_change_sti = rng.choice([2, 3])
+        self.fold_change_w = rng.choice([1., 1.5, 2.], p=[0.05, 0.25, 0.7])
+        self.fold_change_yw = rng.choice([1., 2., 3.]) * self.fold_change_w
+        self.fold_change_sti = rng.choice([2., 3.])
         self.tr_rate_primary = 0.16
         self.tr_rate_undetectable_vl = rng.choice([0.0000, 0.0001, 0.0010], p=[0.7, 0.2, 0.1])
         self.transmission_means = self.fold_tr_newp * np.array([0, self.tr_rate_undetectable_vl, 0.01, 0.03, 0.06, 0.1, self.tr_rate_primary])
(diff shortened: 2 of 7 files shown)


src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
         return hiv_status
 
-    def update_HIV_status(self, population: pd.DataFrame):
+    def update_partner_risk_vectors(self, population):
+        """calculate the risk factor associated with each sex and age group"""
+        # Should we be using for loops here or can we do better? 
+        for sex in SexType:
+            for age_group in range(5):   # FIXME need to get number of age groups from somewhere
+                sub_pop_indices = population.data.index[(population.data[col.SEX]==sex) & (population.data[col.SEX_MIX_AGE_GROUP] == age_group)]
+                sub_pop = population.data.loc[sub_pop_indices]
```
ageorgou: I think we can get this in one step?
```suggestion
                sub_pop = population.data.loc[(population.data[col.SEX]==sex) & (population.data[col.SEX_MIX_AGE_GROUP] == age_group)]
```

src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
+                n_stp_total = sum(sub_pop[col.NUM_PARTNERS])  # total number of people partnered to people in this group
+                n_infected = sum(sub_pop.loc[sub_pop[col.HIV_STATUS]][col.NUM_PARTNERS])  # num people parters to HIV+ people in this group
```
ageorgou: Slightly more idiomatic and (for me) a bit easier to read?
```suggestion
                n_infected = sum(sub_pop.loc[sub_pop[col.HIV_STATUS], col.NUM_PARTNERS])  # num people parters to HIV+ people in this group
```

src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
         hiv_status = self._prob_HIV_initial(population[col.AGE]) > rng.random(len(population))
         return hiv_status
 
-    def update_HIV_status(self, population: pd.DataFrame):
+    def update_partner_risk_vectors(self, population):
+        """calculate the risk factor associated with each sex and age group"""
+        # Should we be using for loops here or can we do better?
```
ageorgou: We *should* be able to do this with a groupby, which would give us the subpopulations directly.

src/hivpy/hiv_status.py
ageorgou: Unless I'm misunderstanding, would this be better called `n_partners_infected` or `n_stp_of_infected` or similar?

src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
+                # Chances of being in a given viral group
+                if n_stp_total > 0:
+                    self.stp_viral_group_rate[sex][age_group] = [sum(sub_pop.loc[sub_pop[col.VIRAL_LOAD_GROUP] == vg, col.NUM_PARTNERS])/n_stp_total for vg in range(7)]
+                else:
+                    self.stp_viral_group_rate[sex][age_group] = np.array([1,0,0,0,0,0,0])
+
+
+    def set_dummy_viral_load(self, population):
+        """Dummy function to set viral load until this part of the code has been implemented properly"""
+        population.data[col.VIRAL_LOAD_GROUP] = rng.choice(7, population.size)
```
ageorgou, line 77: Let's add this variable to #48? Or make an issue somewhere for properly implementing it.

src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
+
+    def stp_HIV_transmission(self, person):
+        # TODO: Add circumcision, STIs etc. 
+        """Returns True if HIV transmission occurs, and False otherwise"""
+        stp_viral_groups = np.array([rng.choice(7, p=self.stp_viral_group_rate[1 - person[col.SEX]][age_group]) for age_group in person[col.STP_AGE_GROUPS]])
```
ageorgou: Instead of the `1 - person[col.SEX]`, we could add a method to the `SexType` enum (or a convenience function) that gives the "opposite" sex.

src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
+        HIV_probabilities = np.array([self.stp_HIV_rate[1 - person[col.SEX]][age_group] for age_group in person[col.STP_AGE_GROUPS]])
```
ageorgou: The SAS code uses "rate" and "probability" almost interchangeably, resulting in some confusion. It'd be worth being more consistent from the start in the new model - if we can understand what each value is...

src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
+        viral_transmission_probabilities = np.array([max(0, rng.normal(self.transmission_means[group], self.transmission_sigmas[group])) for group in stp_viral_groups])
+        if person[col.SEX] == SexType.Female:
```
ageorgou: Preferred per [the docs on enums](https://docs.python.org/3/library/enum.html#comparisons):
```suggestion
        if person[col.SEX] is SexType.Female:
```

src/hivpy/hiv_status.py
```diff
@@ -24,17 +39,54 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
-        rands = rng.uniform(0.0, 1.0, sum(HIV_neg_idx))
-        HIV_prevalence = sum(population[col.HIV_STATUS])/len(population)
-        HIV_infection_risk = 0.1  # made up, based loosely on transmission probabilities
-        n_partners = population.loc[HIV_neg_idx, col.NUM_PARTNERS]
-        HIV_prob = 1-((1-HIV_prevalence*HIV_infection_risk)**n_partners)
-        population.loc[HIV_neg_idx, col.HIV_STATUS] = (rands <= HIV_prob)
+        self.update_partner_risk_vectors(population)
+        HIV_neg_idx = population.data.index[(population.data[col.HIV_STATUS]==False) & (population.data[col.NUM_PARTNERS]>0)]
+        sub_pop = population.data.loc[HIV_neg_idx]
+        population.data.loc[HIV_neg_idx, col.HIV_STATUS] = sub_pop.apply(self.stp_HIV_transmission, axis=1)
```
ageorgou: Hm, this seems like a prime candidate for vectorization if we can. But I guess the value of `STP_AGE_GROUPS` will be different for everyone? Perhaps we could combine the two steps (choice of partner age groups and probability of transmission) somehow.

ageorgou (commented):
I've not read through the whole thing yet, only making some quick comments.

I need to look at the transmission probabilities more closely because it seems we get the ratios the wrong way around, although that's probably me reading it wrong!

Does applying `stp_HIV_transmission` to each person individually give okay performance? (from what we can tell, anyway)

We may need to rethink the age groups and if they should be in `common`, like the sex type. Otherwise pass a reference to the population into the module?

src/hivpy/hiv_status.py
mmcleod89: So I thought about this using a `groupby` but I don't know if it can be done because we need to access HIV status and  number of partners, but we don't want to group by these, only by sex and age group. I'm not sure how we can access these additional fields if we use a group by

src/hivpy/hiv_status.py
mmcleod89: Yes you're correct, that name should have been changed when I corrected this calculation (I mistakenly used the number of people with HIV at first instead of the number of people coupled to someone with HIV) so will change this

src/hivpy/hiv_status.py
mmcleod89: I very much forgot #48 existed, so yes I will add this to it.

src/hivpy/hiv_status.py
mmcleod89: This is a good point; I suppose the most correct way to use it would be to use rate for things which need to be time-step adjusted, so I which variables you call _rate_ or _probability_ will depend on at which point in the calculation we take the length of the timestep into account.

mmcleod89: At the moment this doesn't take timestep length into account at all, so there isn't really a difference between the notion of a rate or a probability yet

src/tests/test_hiv_status.py
```diff
@@ -61,40 +62,45 @@ def test_HIV_risk_vector():
...
     assert np.allclose(hiv_module.stp_HIV_rate[SexType.Female], expectation)
 
     # Check for differences in male and female rate correctly
-    # change HIV rate in men to double 
+    # change HIV rate in men to double
     males = pop.data.index[pop.data[col.SEX] == SexType.Male]
-    # transform group fails when only grouped by one field --> appears to change the type of the object passed to the function! 
-    male_HIV_status = pop.transform_group([col.SEX_MIX_AGE_GROUP, col.SEX], lambda x, y: np.array([True] * (2 * N_group // HIV_ratio) + [False] * (N_group - 2*N_group // HIV_ratio)), False, males)
+    # transform group fails when only grouped by one field
+    # appears to change the type of the object passed to the function!
```
ageorgou, line 137: Just noting that this should be fixed as part of #71, so we can update the following lines when merged.

src/tests/test_hiv_status.py
```diff
@@ -61,40 +62,45 @@ def test_HIV_risk_vector():
...
+    male_HIV_status = pop.transform_group([col.SEX_MIX_AGE_GROUP, col.SEX], lambda x, y: np.array(
+        [True] * (2 * N_group // HIV_ratio) +
+        [False] * (N_group - 2*N_group // HIV_ratio)), False, males)
```
ageorgou, line 140: Although I'm confused by what the (lambda) function does here: neither argument is used?

src/tests/test_hiv_status.py
```diff
@@ -140,9 +148,11 @@ def test_hiv_update2():
...
+    expected_transmission_prob = 0.1 * 0.1
     pop.data["delta_HIV"] = pop.data[col.HIV_STATUS] ^ prev_hiv_status
-    n_new_male = sum(pop.data.loc[pop.data[col.SEX]==SexType.Male, "delta_HIV"])  # count number of new transmissions for men 
+    # count number of new transmissions for men
+    n_new_male = sum(pop.data.loc[pop.data[col.SEX] == SexType.Male, "delta_HIV"])
     exp_new_male = 45000 * expected_transmission_prob
     sig = np.sqrt(45000 * expected_transmission_prob * (1 - expected_transmission_prob))
-    print(n_new_male, exp_new_male, sig)
\ No newline at end of file
+    print(n_new_male, exp_new_male, sig)
```
ageorgou: Remove when done :grin:

src/hivpy/common.py
```diff
@@ -35,6 +35,8 @@ class SexType(IntEnum):
     Male = 0
     Female = 1
 
+def opposite_sex(sex: SexType):
+    return (1 - sex)
```
ageorgou, line 41: Personally I would prefer to keep the abstraction here, so something like
```python
return SexType.Male if sex is SexType.Female else SexType.Female
```

src/hivpy/hiv_status.py
```diff
@@ -64,12 +62,31 @@ def set_dummy_viral_load(self, population):
...
+        for i in range (n_partners):
+            stp_viral_group = rng.choice(7, p = self.stp_viral_group_rate[target_sex][stp_age_groups[i]])
+            HIV_probability = self.stp_HIV_rate[opposite_sex(target_sex)][stp_age_groups[i]]
+            infection_prob[i] = HIV_probability * max(0, rng.normal(self.transmission_means[stp_viral_group], self.transmission_sigmas[stp_viral_group]))
+            if (sex == SexType.Female):
+                if(age < 20):
+                    infection_prob[i] *= self.fold_change_yw
+                else:
+                    infection_prob[i] *= self.fold_change_w
+            return infection_prob
```
ageorgou: This should be out of the loop? (currently returning after computing probability for the first partner only?)

src/hivpy/hiv_status.py
```diff
@@ -64,12 +62,31 @@ def set_dummy_viral_load(self, population):
         """Dummy function to set viral load until this part of the code has been implemented properly"""
         population.data[col.VIRAL_LOAD_GROUP] = rng.choice(7, population.size)
 
+    def get_infection_prob(self, sex, age, n_partners, stp_age_groups):
+        # Slow example that avoid repeating the iterations over partners three time by putting them as part of 
+        # one for loop, but for loops in python will be slow.
```
ageorgou: Yeah, it'd be nice to be able to vectorise this, but having to repeat it for multiple partners makes it tricky... Will try to think of something for a future version.

src/hivpy/common.py
ageorgou: Ah, but that doesn't work if called on a whole column, does it? Hm. There's probably a way to vectorise this while preserving the abstraction but I guess it's not crucial.

src/hivpy/hiv_status.py
```diff
@@ -24,17 +40,87 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
+    def update_partner_risk_vectors(self, population):
+        """calculate the risk factor associated with each sex and age group"""
+        # Should we be using for loops here or can we do better?
+        for sex in SexType:
+            for age_group in range(5):   # FIXME need to get number of age groups from somewhere
+                sub_pop = population.data.loc[(population.data[col.SEX] == sex) & (
+                    population.data[col.SEX_MIX_AGE_GROUP] == age_group)]
+                # total number of people partnered to people in this group
+                n_stp_total = sum(sub_pop[col.NUM_PARTNERS])
+                # num people partered to HIV+ people in this group
```
ageorgou, line 58:
```suggestion
                # num people partnered to HIV+ people in this group
```

src/hivpy/hiv_status.py
```diff
@@ -24,17 +40,87 @@ def initial_HIV_status(self, population: pd.DataFrame):
...
+                    infection_prob[i] *= self.fold_change_yw
+                else:
+                    infection_prob[i] *= self.fold_change_w
+            return infection_prob
+
+    def stp_HIV_transmission(self, person):
+        # TODO: Add circumcision, STIs etc.
+        """Returns True if HIV transmission occurs, and False otherwise"""
+        stp_viral_groups = np.array([rng.choice(7, p=self.stp_viral_group_rate[opposite_sex(
+            person[col.SEX])][age_group]) for age_group in person[col.STP_AGE_GROUPS]])
```
ageorgou: I find something like the following a more readable way to break it up:
```suggestion
        stp_viral_groups = np.array([
            rng.choice(7, p=self.stp_viral_group_rate[opposite_sex(person[col.SEX])][age_group])
            for age_group in person[col.STP_AGE_GROUPS]
        ])
```
but that's personal preference! (and the second line is actually longer this way... though less than 100)

ageorgou (changes requested):
I think there's a bug in the infection probabilities, unless I'm misunderstanding - otherwise looks fine though!

src/tests/test_hiv_status.py
mmcleod89: This is just because the function passed to transform group has to match the number of arguments to the variables grouped by, which is something we can potentially change by making transform_group more flexible

src/tests/test_hiv_status.py
mmcleod89: This whole test was already removed in [a62a146](https://github.com/UCL/hivpy/pull/70/commits/a62a1461fab2b11dc63fba38253f8590a642f28a)

src/hivpy/hiv_status.py
mmcleod89: Yes, it should be out of the loop !

mmcleod89:
test_rred keeps failing on this and another branch but not on my machine; is this because of the pandas issue we saw before?

ageorgou:
No, it's a pandas bug in 1.5.0, see also [this comment](https://github.com/UCL/hivpy/pull/74#issuecomment-1264630383). We're pinning it to 1.3.x in #71 but haven't added that here.

ageorgou (approved):

PR #71, opened 2022-08-22
WIP: Add long term partners

mmcleod89:
Needs tests and some balancing factors for age / sex

### DIFF
--- a/src/hivpy/sexual_behaviour.py
+++ b/src/hivpy/sexual_behaviour.py
@@ -332,17 +332,18 @@ class SexualBehaviourModule:
     def new_ltp_longevity(self, age_group, size):
         longevity = np.zeros(size)  # each new relationship needs a longevity
         longevity_rands = rng.random(size)
+        thresholds = [0, 0]
         if age_group < 2:
-            longevity = ((longevity_rands < 0.3) * 1 +
-                         ((0.3 <= longevity_rands) & (longevity_rands < 0.6)) * 2 +
-                         (0.6 <= longevity_rands) * 3)
+            thresholds = [0.3, 0.6]
         elif age_group == 2:
-            longevity = ((longevity_rands < 0.3) * 1 +
-                         ((0.3 <= longevity_rands) & (longevity_rands < 0.8)) * 2 +
-                         (0.8 <= longevity_rands) * 3)
+            thresholds = [0.3, 0.8]
         else:
-            longevity = ((longevity_rands < 0.3) * 1 +
-                         (0.3 <= longevity_rands) * 2)
+            thresholds = [0.3, 1.0]
+
+        longevity[longevity_rands < thresholds[0]] = 1
+        longevity[(thresholds[0] <= longevity_rands) & (longevity_rands < thresholds[1])] = 2
+        longevity[thresholds[1] <= longevity_rands] = 3
+
         return longevity
 
     def continue_ltp(self, longevity, size):
--- a/src/hivpy/sexual_behaviour.py
+++ b/src/hivpy/sexual_behaviour.py
@@ -80,6 +80,7 @@ class SexualBehaviourModule:
         self.ltp_rate_change = 1.0
         self.ltp_end_rate_by_longevity = np.array([0, 0.25, 0.05, 0.02])
         self.ltp_balance_factor = {SexType.Male: 1, SexType.Female: 1}
+        self.new_ltp_age_bins = [35, 45, 55]
 
     def age_index(self, age):
         return np.minimum((age.astype(int)-self.risk_min_age) //
@@ -313,29 +314,29 @@ class SexualBehaviourModule:
         else:
             return (1, 1)
 
-    def start_new_ltp(self, sex, age, size):
+    def start_new_ltp(self, sex, age_group, size):
         # TODO add alterations for new diagnosis
         ltp_rands = rng.random(size=size)
         rate_modifier = 1
-        if (35 <= age < 45):
+        if age_group == 1:
             rate_modifier = 0.5
-        elif (45 <= age < 55):
+        elif age_group == 2:
             rate_modifier = 1/3
-        elif (55 <= age):
+        elif age_group == 3:
             rate_modifier = 0.2
         new_relationship = ltp_rands < (self.new_ltp_rate
                                         * rate_modifier
                                         * self.ltp_balance_factor[sex])
         return new_relationship
 
-    def new_ltp_longevity(self, age, sex, size):
+    def new_ltp_longevity(self, age_group, size):
         longevity = np.zeros(size)  # each new relationship needs a longevity
         longevity_rands = rng.random(size)
-        if (age < 45):
+        if age_group < 2:
             longevity = ((longevity_rands < 0.3) * 1 +
                          ((0.3 <= longevity_rands) & (longevity_rands < 0.6)) * 2 +
                          (0.6 <= longevity_rands) * 3)
-        elif (age < 55):
+        elif age_group == 2:
             longevity = ((longevity_rands < 0.3) * 1 +
                          ((0.3 <= longevity_rands) & (longevity_rands < 0.8)) * 2 +
                          (0.8 <= longevity_rands) * 3)
@@ -364,17 +365,20 @@ class SexualBehaviourModule:
 
         (self.ltp_balance_factor[SexType.Male],
          self.ltp_balance_factor[SexType.Female]) = self.balance_ltp_factor(population)
+
         # new relationships
+        population.data[col.LTP_AGE_GROUP] = np.digitize(
+            population.data[col.AGE], self.new_ltp_age_bins)
         new_relationships = population.transform_group(
-            [col.SEX, col.AGE],  self.start_new_ltp, sub_pop=partnerless_idx)
+            [col.SEX, col.LTP_AGE_GROUP],  self.start_new_ltp, sub_pop=partnerless_idx)
         population.data.loc[partnerless_idx, col.LONG_TERM_PARTNER] = new_relationships
-        # not sure about indexing on this next line
-        # We only want to update LTP_LONGEVITY where LONG_TERM_PARTNER is true within
-        # the sub-population defined by partnerless_idx
+
+        # calculate longevity for new relationships this time step
+        # (but not for existing relationships)
         new_ltp_subpop = population.data.index[start_partnerless &
                                                population.data[col.LONG_TERM_PARTNER]]
         longevity = population.transform_group(
-            [col.AGE, col.SEX], self.new_ltp_longevity, sub_pop=new_ltp_subpop)
+            [col.LTP_AGE_GROUP], self.new_ltp_longevity, sub_pop=new_ltp_subpop)
         population.data.loc[new_ltp_subpop, col.LTP_LONGEVITY] = longevity.astype(int)
         # ending relationships
         population.data.loc[partnered_idx, col.LONG_TERM_PARTNER] = population.transform_group(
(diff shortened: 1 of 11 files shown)


src/hivpy/sexual_behaviour.py
```diff
@@ -74,6 +74,13 @@ def __init__(self, **kwargs):
         self.balance_factors = [0.1, 0.7, 0.7, 0.75, 0.8, 0.9, 0.97]
         self.p_rred_p = self.sb_data.p_rred_p_dist.sample()
 
+        # long term partnerships
+        self.new_ltp_rate = 0.1 * np.exp(rng.normal() * 0.25)  # three month ep-rate
+        self.annual_ltp_rate_change = rng.choice([0.8, 0.9, 0.95, 1.0])
+        self.ltp_rate_change = 1.0
+        self.ltp_end_rate_by_longevity = np.array([0, 0.25, 0.05, 0.02])
```
ageorgou, line 81: These could go in the data file?

src/hivpy/sexual_behaviour.py
```diff
@@ -277,3 +284,99 @@ def update_rred_balance(self, population):
...
             population.loc[men, col.RRED_BALANCE] = 1/rred_balance
             population.loc[women, col.RRED_BALANCE] = rred_balance
+
+    def update_ltp_rate_change(self, date):
+        if date1995 < date < date2000:
+            dt = diff_years(date1995, date)
+            self.ltp_rate_change = self.annual_ltp_rate_change**(dt)
+        elif date >= date2000:
+            self.ltp_rate_change = self.annual_ltp_rate_change**5
+            # don't really need to keep doing this update after 2000
```
ageorgou, line 295: There is something similar in the HIV_status code, and there will probably be more. One idea (for the future!) could be to track the date only in the Population, and have modules "subscribe" to date changes, so they can update their internal parameters when needed (and ignore any date change after that).

src/hivpy/sexual_behaviour.py
```diff
@@ -277,3 +284,99 @@ def update_rred_balance(self, population):
...
+    def start_new_ltp(self, sex, age, size):
+        # TODO add alterations for new diagnosis
+        ltp_rands = rng.random(size=size)
+        rate_modifier = 1
+        if (35 <= age < 45):
+            rate_modifier = 0.5
+        elif (45 <= age < 55):
+            rate_modifier = 1/3
+        else:
+            rate_modifier = 0.2
```
ageorgou: Could initialise to 0.2 or remove the first line?
Or a possibly over-complicated
```suggestion
        rate_modifier = 0.5 if 35 <= age < 45 else 1/3 if 45 <= age < 55 else 0.2
```

src/hivpy/sexual_behaviour.py
```diff
@@ -277,3 +284,99 @@ def update_rred_balance(self, population):
...
+        new_relationship = ltp_rands < (self.new_ltp_rate
+                                        * rate_modifier
+                                        * self.ltp_balance_factor[sex])
+        return new_relationship
+
+    def new_ltp_longevity(self, age, sex, size):
```
ageorgou: Minor, but change the argument orders for consistency with `start_new_ltp`? (either way seems fine!)

src/hivpy/sexual_behaviour.py
```diff
@@ -277,3 +284,99 @@ def update_rred_balance(self, population):
...
+        longevity = np.zeros(size)  # each new relationship needs a longevity
+        longevity_rands = rng.random(size)
+        if (age < 45):
+            longevity = ((longevity_rands < 0.3) * 1 +
+                         ((0.3 <= longevity_rands) & (longevity_rands < 0.6)) * 2 +
+                         (0.6 <= longevity_rands) * 3)
```
ageorgou: My initial thought about writing this would be with `.loc` - any benefits to either?
  ```suggestion
              longevity.loc[longevity_rands < 0.3, :] = 1
              longevity.loc[(0.3 <= longevity_rands) & (longevity_rands < 0.6), :] = 2
              longevity.loc[0.6 <= longevity_rands, :] = 3
```

Comment (ageorgou) [reply src/hivpy/sexual_behaviour.py]: I also wonder if combining the code for different age categories makes sense, e.g.
```python
cat2_threshold = 0.6 if age < 45 else 0.8 if age < 55 else 1.1
...
longevity.loc[(0.3 <= longevity_rands) & (longevity_rands < cat2_threshold), :] = 2
```
(or with the sum approach as is)

src/hivpy/sexual_behaviour.py
```diff
@@ -277,3 +284,99 @@ def update_rred_balance(self, population):
...
+        else:
+            longevity = ((longevity_rands < 0.3) * 1 +
+                         (0.3 <= longevity_rands) * 2)
+        return longevity
+
+    def continue_ltp(self, age, longevity, size):
+        """Function to decide which long term partners cease condomless sex based on
+        relationship longevity, age, and sex."""
+        # TODO: Add balancing factors for age and sex demographics.
+        end_probability = self.ltp_end_rate_by_longevity[int(longevity)] / self.ltp_rate_change
```
ageorgou: Is there any chance longevity is not an integer?

src/hivpy/sexual_behaviour.py
```diff
@@ -277,3 +284,99 @@ def update_rred_balance(self, population):
...
+
+        # new relationships
+        new_relationships = population.transform_group(
+            [col.SEX, col.AGE],  self.start_new_ltp, sub_pop=partnerless_idx)
+        population.data.loc[partnerless_idx, col.LONG_TERM_PARTNER] = new_relationships
+        # not sure about indexing on this next line
+        # We only want to update LTP_LONGEVITY where LONG_TERM_PARTNER is true within
+        # the sub-population defined by partnerless_idx
+        new_ltp_subpop = population.data.index[start_partnerless &
+                                               population.data[col.LONG_TERM_PARTNER]]
```
ageorgou: The indexing looks okay to me: they were partnerless at the start and now have a partner

src/hivpy/sexual_behaviour.py
```diff
@@ -277,3 +284,99 @@ def update_rred_balance(self, population):
...
+        partnered_idx = population.data.index[(population.data[col.LONG_TERM_PARTNER]) &
+                                              (population.data[col.AGE] >= 15) &
+                                              (population.data[col.AGE] < 65)]
+
+        (self.ltp_balance_factor[SexType.Male],
+         self.ltp_balance_factor[SexType.Female]) = self.balance_ltp_factor(population)
+
+        # new relationships
+        new_relationships = population.transform_group(
+            [col.SEX, col.AGE],  self.start_new_ltp, sub_pop=partnerless_idx)
```
ageorgou: One thing to note here and for longevity is that age is a continuous column (well, discretised by the time step, but not integer) so the grouping may be inefficient. Would doing it by age group be okay? (though we lose the clarity of the explicit age in the code)

src/hivpy/sexual_behaviour.py
ageorgou: Oh, sorry, `longevity` is a numpy array, not a panda object. We can still assign a subset of elements with a logical index though:
```python
longevity[longevity_rands < 0.3] = 1
```

ageorgou:
This all looks correct, just a few suggestions above. The tests for starting/ending relationships also look reasonable, though they may be better as parameterised tests instead of running the loops inside them. I think the failures are down to HIV infections not always appearing, which is an issue I also had in #69, even after altering the rates to make it it easier to pass...

src/hivpy/sexual_behaviour.py
mmcleod89: There shouldn't be but `pandas` is sometimes insane and converts it to a float somehow and this is not at all predictable as far as I can yet tell (*cough* something about type systems *cough*)

src/hivpy/sexual_behaviour.py
```diff
@@ -313,38 +314,39 @@ def balance_ltp_factor(self, population):
...
         rate_modifier = 1
-        if (35 <= age < 45):
+        if age_group == 1:
             rate_modifier = 0.5
-        elif (45 <= age < 55):
+        elif age_group == 2:
             rate_modifier = 1/3
-        elif (55 <= age):
+        elif age_group == 3:
             rate_modifier = 0.2
```
ageorgou: Another way of writing it with a dictionary (and a default value for other age groups):
```python
rate_modifiers_per_group = {1: 0.5, 2: 1/3, 3: 0.2}
rate_modifier = rate_modifiers_per_group.get(age_group, 1)
```

src/hivpy/sexual_behaviour.py
```diff
@@ -313,38 +314,39 @@ def balance_ltp_factor(self, population):
...
         longevity_rands = rng.random(size)
-        if (age < 45):
-            longevity = ((longevity_rands < 0.3) * 1 +
-                         ((0.3 <= longevity_rands) & (longevity_rands < 0.6)) * 2 +
-                         (0.6 <= longevity_rands) * 3)
-        elif (age < 55):
-            longevity = ((longevity_rands < 0.3) * 1 +
-                         ((0.3 <= longevity_rands) & (longevity_rands < 0.8)) * 2 +
-                         (0.8 <= longevity_rands) * 3)
+        thresholds = [0, 0]
```
ageorgou, line 335: This default value is always overwritten, right?

src/hivpy/sexual_behaviour.py
```diff
@@ -364,18 +366,21 @@ def update_long_term_partners(self, population):
...
                                                population.data[col.LONG_TERM_PARTNER]]
         longevity = population.transform_group(
-            [col.AGE, col.SEX], self.new_ltp_longevity, sub_pop=new_ltp_subpop)
-        population.data.loc[new_ltp_subpop, col.LTP_LONGEVITY] = longevity
+            [col.LTP_AGE_GROUP], self.new_ltp_longevity, sub_pop=new_ltp_subpop)
+        population.data.loc[new_ltp_subpop, col.LTP_LONGEVITY] = longevity.astype(int)
         # ending relationships
         population.data.loc[partnered_idx, col.LONG_TERM_PARTNER] = population.transform_group(
-            [col.AGE, col.LTP_LONGEVITY], self.continue_ltp, sub_pop=partnered_idx)
+            [col.LTP_LONGEVITY], self.continue_ltp, sub_pop=partnered_idx)
```
ageorgou: Taking this as an opportunity to highlight that this may be worth coding as a higher-level method on the `Population`: it appears twice in this method alone, and I think it would be more readable. Something like
```python
population.update_column_by_groups(
    column_name=col.LONG_TERM_PARTNER,
    group_by=[col.LTP_LONGEVITY],
    apply=self.continue_ltp,
    sub_pop=partnered_idx
)
```
that calls `transform_group` and assigns the result to the specified column.

src/tests/test_hiv_status.py
```diff
@@ -27,11 +27,11 @@ def test_hiv_update():
     pop = Population(size=pop_size, start_date=date(1989, 1, 1))
     data = pop.data
     prev_status = data["HIV_status"].copy()
-
     for i in range(10):
         pop.hiv_status.update_HIV_status(pop.data)
 
     new_cases = data["HIV_status"] & (~ prev_status)
+    print(sum(new_cases))
```
ageorgou: Delete this when ready? :grin:

src/hivpy/sexual_behaviour.py
mmcleod89: This default value is always overwritten because all cases are covered below; I just really don't like the thing of declaring variables inside conditional logic which then need to be used outside of it

src/tests/test_hiv_status.py
mmcleod89: Oops! I thought I had got rid of those!

ageorgou (approved):

PR #74, opened 2022-09-30
Basic guide and some tidying up

ageorgou:
- Filled in the README with some context and instructions
- Added a license (chose MIT but don't have particular concerns)
- Removed an earlier attempt at a second command that was not being used anywhere

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,9 @@ This will install some additional libraries and tools for development, and also 
 
 ### Testing
 The package comes with unit tests. To run them, simply run
-`pytest` from the top level (this directory). The `pytest` package and command are installed during 
+`pytest` from the top level (this directory). The `pytest` package and command
+are installed during the package installation if the `[tests]` extra option
+is supplied, as above.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1 +1,95 @@
...
+If you are planning to make changes to the package code,
+use the following command to install the package instead:
+```bash
+pip install -e .[tests]
+```
+This will install some additional libraries and tools for development, and also create an "editable" installation; this means that any changes you make will be applied automatically, without needing to reinstall the package.
+
+### Testing
+The package comes with unit tests. To run them, simply run
+`pytest` from the top level (this directory). The `pytest` package and command are installed during
```
pineapple-cat: [...] installed during [the pip install]?

ageorgou:
The tests are failing because of [a bug in pandas 1.5.0](https://github.com/pandas-dev/pandas/issues/48701), which should be fixed in the next release.

pineapple-cat (approved):

PR #77, opened 2022-10-04
Fix minor issues

### LINKED ISSUE #75: Empy column header in output file
The file with simulation outputs has no header for the first column (increasing id of time step). Reading it as CSV still works, but it would be nicer to show something.

### LINKED ISSUE #76: Create log folder if necessary
If the code is run without a log folder present in the expected place, it will try and fail to create a log file because it is not creating the full path. 

If there is no log folder (or other missing folders along the path to the log folder) one needs to be created.

pineapple-cat:
Solves issues #75 and #76.

### DIFF
--- a/src/hivpy/experiment.py
+++ b/src/hivpy/experiment.py
@@ -71,9 +71,10 @@ def create_simulation(experiment_param):
 
 
 def create_log(log_param):
-    log_dir = Path(log_param['log_directory'])
-    if not log_dir.exists():
-        log_dir.mkdir(parents=True)
+    log_dir = log_param['log_directory']
+    log_dir_path = Path(log_dir)
+    if not log_dir_path.exists():
+        log_dir_path.mkdir(parents=True)
     log_file = log_param['logfile_prefix']+"_"+datetime.now().strftime("%y%m%d-%H%M%S")+".log"
     log_level = log_param['log_file_level']
     console_log_level = log_param['console_log_level']
--- a/src/hivpy/experiment.py
+++ b/src/hivpy/experiment.py
@@ -72,7 +72,7 @@ def create_simulation(experiment_param):
 
 def create_log(log_param):
     log_dir = Path(log_param['log_directory'])
-    if not log_dir.exists(): 
+    if not log_dir.exists():
         log_dir.mkdir(parents=True)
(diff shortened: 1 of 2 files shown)


src/hivpy/experiment.py
```diff
@@ -71,7 +71,9 @@ def create_simulation(experiment_param):
 
 
 def create_log(log_param):
-    log_dir = log_param['log_directory']
+    log_dir = Path(log_param['log_directory'])
```
ageorgou: Note that this means we pass a `Path` rather than a string to the `LoggingConfig` constructor, which contradicts the type hints for that argument. But it still works fine!

ageorgou (approved):
Looks good to me. The tests are failing as in #74 because of a pandas bug, so I think this can be merged.

Is it worth harmonising the code in `config.py` to only use one of `os.path` or `pathlib.Path`? Although they're both fully supported, so I don't think this is particularly important.

src/hivpy/experiment.py
pineapple-cat: I was wondering whether it was worth storing the path separately, but I tried it both ways and it worked the same so I figured it must be fine. (Famous last words..?) I could also `p.as_posix()` the path when it is passed to `LoggingConfig`, just in case.

ageorgou: It is fine in practice: `os.path.join` (which is what this `log_dir` is eventually passed to inside `LoggingConfig`) will accept any [path-like argument](https://docs.python.org/3/library/os.html#os.PathLike), whether it's a string or a `Path` object.

This is only introducing a minor inconsistency in the type hints of `LoggingConfig`, where we say that `log_dir` is a string. I think the easiest way around it is using an extra local variable for the `Path` inside `create_log` (and for creating the parents if needed), and passing in the string itself to `LoggingConfig`.

Alternatively, we could:
- change the type hint of `log_dir` in `LoggingConfig` to something more general, as suggested in [this answer](https://stackoverflow.com/questions/53418046/how-do-i-type-hint-a-filename-in-a-function)
- or switch over everything to use `Path`s instead of `os.path` functions, but that's more work!

pineapple-cat: Easy way it is!

PR #78, opened 2022-10-06
Basic template for Sphinx

ageorgou:
Including new files created with `sphinx-quickstart` and `sphinx-apidoc`, with some changes. Also adding new dependency for sphinx for the "development" install (which we should perhaps rename from `tests`).

Docs can be built by running `make html` from the `src/docs` directory. The created pages are then in `src/docs/_build`.

If someone can verify that this works, we can add further documentation (either in the code or in the docs pages themselves) in separate PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ output/
 *.egg
 __pycache__/
 .pytest_cache/
+
+# Generated documentation
+_build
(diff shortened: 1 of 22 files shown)


mmcleod89 (approved):
Don't know much about sphinx yet, but looks okay to me and works when I try to build!

pineapple-cat (approved):
Build works for me too, but not with Git Bash for some reason. Still trying to figure out why, but it's definitely a me problem.

PR #92, opened 2022-10-18
Add circumcision feature

### LINKED ISSUE #81: Circumcision

### LINKED ISSUE #94: Circumcision Implementation & Timing
Circumcision at birth implemented currently for entire population (born and unborn). 
Can be timed and then compared to the implementation where there are two stages: circumcision at birth for the initial population male and age > 0, and then the update circumcision at every time step. 

Can use `perf_counter` from the `time` library to time functions. 
Can time circumcision update function for multiple function calls and divide the time to get the average to get a more accurate result. 

Likely that impact on performance will be small anyway.

pineapple-cat:
To implement:

- [x] Circumcision at birth
- [x] VMMC
- [x] Change in VMMC over time
- [x] Change in VMMC due to age
- [x] COVID disruption (dummy)


pineapple-cat:
Encountered and fixed a bug when writing `update_vmmc` involving an unusual Empty DataFrame case.

**Initial VMMC code:**

```
def update_vmmc(self, pop):

    self.date = pop.date
    if self.mc_int < self.date.year:

        # get uncirc males
        uncirc_male_population = pop.data.index[(pop.data[col.SEX] == SexType.Male)
                                                & ~pop.data[col.CIRCUMCISED]
                                                & (pop.data[col.AGE] >= 10)
                                                & (pop.data[col.AGE] < 50)]

        # group males by age group
        pop.data.loc[uncirc_male_population, 
                     col.AGE_GROUP] = np.digitize(pop.data.loc[uncirc_male_population, col.AGE],
                                                               [10, 20, 30, 50])

        # calculate circ outcomes
        circumcision = pop.transform_group([col.AGE_GROUP], self.calc_prob_circ, 
                                           sub_pop=uncirc_male_population)

        # apply circ
        pop.data.loc[uncirc_male_population, col.CIRCUMCISED] = circumcision
        pop.data.loc[uncirc_male_population, col.VMMC] = circumcision

        # set date
        new_circ_males = pop.data.index[pop.data[col.CIRCUMCISED]
                                        & pop.data[col.CIRCUMCISION_DATE].isnull()]
        pop.data.loc[new_circ_males, col.CIRCUMCISION_DATE] = self.date
```

**Initial test results from trying to replicate the issue:**

```
New Artificial Population:
    sex  age  circumcised circumcision_date   vmmc
0    0   12        False              None  False
1    0   12        False              None  False
2    0   12        False              None  False
3    0   12        False              None  False
4    0   12        False              None  False
```

===== EVOLVE STEP 1 =====
Normal step. Behaviour as expected. Circumcises all males (circumcision probability was set to 100% for the test case).

```
Found Uncircumcised Male Population:
    sex        age  age_group  circumcised circumcision_date   vmmc
0    0  12.246575        1.0        False              None  False
1    0  12.246575        1.0        False              None  False
2    0  12.246575        1.0        False              None  False
3    0  12.246575        1.0        False              None  False
4    0  12.246575        1.0        False              None  False
```

```
Circumcision Outcomes:
0    True
1    True
2    True
3    True
4    True
```

```
Population After Evolve #1:
    sex        age  circumcised circumcision_date  vmmc
0    0  12.246575         True        2009-01-01  True
1    0  12.246575         True        2009-01-01  True
2    0  12.246575         True        2009-01-01  True
3    0  12.246575         True        2009-01-01  True
4    0  12.246575         True        2009-01-01  True
```

===== EVOLVE STEP 2 =====
Normal step. No uncircumcised males are discovered. Nobody is circumcised.

```
Found Uncircumcised Male Population:
Empty DataFrame
```

```
Circumcision Outcomes:
Series([], dtype: float64)
```

```
Population After Evolve #2:
    sex        age circumcised circumcision_date  vmmc
0    0  12.493151        True        2009-01-01  True
1    0  12.493151        True        2009-01-01  True
2    0  12.493151        True        2009-01-01  True
3    0  12.493151        True        2009-01-01  True
4    0  12.493151        True        2009-01-01  True
```

===== EVOLVE STEP 3 =====
Problem step. Circumcised males are selected for the uncircumcised population. Everyone is re-circumcised (though circumcision dates are not updated).

```
Found Uncircumcised Male Population:
    sex        age  age_group circumcised circumcision_date  vmmc
0    0  12.739726          1        True        2009-01-01  True
1    0  12.739726          1        True        2009-01-01  True
2    0  12.739726          1        True        2009-01-01  True
3    0  12.739726          1        True        2009-01-01  True
4    0  12.739726          1        True        2009-01-01  True
```

```
Circumcision Outcomes:
0    True
1    True
2    True
3    True
4    True
```

```
Population After Evolve #3:
    sex        age  circumcised circumcision_date  vmmc
0    0  12.739726         True        2009-01-01  True
1    0  12.739726         True        2009-01-01  True
2    0  12.739726         True        2009-01-01  True
3    0  12.739726         True        2009-01-01  True
4    0  12.739726         True        2009-01-01  True
```

Running the test also produced the following warning in several places:

> FutureWarning: In a future version, `df.iloc[:, i] = newvals` will attempt to set the values inplace instead of always setting a new array. To retain the old behavior, use either `df[df.columns[i]] = newvals` or, if columns are non-unique, `df.isetitem(i, newvals)`

**The fix:**

Added an `if len(uncirc_male_population) > 0` statement after `uncirc_male_population` is found and put the rest of the function code behind it, which produced the desired behaviour in step 3.

===== [FIXED] EVOLVE STEP 3 =====

```
Found Uncircumcised Male Population:
Empty DataFrame
```

```
Population After Evolve #3:
    sex        age  circumcised circumcision_date  vmmc
0    0  12.739726         True        2009-01-01  True
1    0  12.739726         True        2009-01-01  True
2    0  12.739726         True        2009-01-01  True
3    0  12.739726         True        2009-01-01  True
4    0  12.739726         True        2009-01-01  True
```

**The cause:**

The unexpected behaviour occurred in pandas 1.5.0 and seemed to be the result of a change in behaviour between pandas versions. HIVpy is currently pandas <1.4.0 compatible, but creating more pandas wrappers and investigating pandas behaviour across the code would help potentially make HIVpy compatible with newer pandas versions as well.

pineapple-cat:
**Note**: The circumcision module is now largely implemented, however there are still some small missing details, such as interactions with the `hard_reach`, `registd`, and `tested` variables.

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+import numpy as np
+
+import hivpy.column_names as col
+from hivpy.common import SexType
+from hivpy.population import Population
+
+
+# Find average birth circumcicion time: init all at once vs
+# partial init followed by average time of updates per time step
+def test_birth_circumcision_timing():
```
mmcleod89: I think as long as we have the results of the timing recorded, we don't need to include the timing in the unit tests right now as it's not something that's likely to change as we develop and we don't want to necessarily run it every time as it takes just under a minute on my machine, a considerable fraction of the total test time so far

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+    # build population
+    N = 100000
+    pop = Population(size=N, start_date=date(1990, 1, 1))
+    pop.data[col.CIRCUMCISED] = False
+    pop.data[col.CIRCUMCISION_DATE] = None
+    pop.data[col.VMMC] = False
+    pop.circumcision.init_birth_circumcision_all(pop.data, pop.date)
+
+    # get stats
+    no_circumcised = len(pop.data[pop.data[col.CIRCUMCISED] == 1])
```
mmcleod89, line 96:
```suggestion
    no_circumcised = sum(pop.data[col.CIRCUMCISED])```

This might be a bit quicker as it only needs to make one pass over the vector and doesn't need to construct a new one.

File: src/tests/test_circumcision.py [hunk 3]
```diff
@@ -0,0 +1,435 @@
...
+    no_male = len(pop.data[pop.data[col.SEX] == SexType.Male])
```
mmcleod89:
```suggestion
    no_circumcised = sum(pop.data[col.CIRCUMCISED])
    no_male = sum(pop.data[col.SEX] == SexType.Male)
```
This might be a little simpler and could save some time as it doesn't need to make a new array

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+    assert len(data[(data[col.SEX] == SexType.Female)
+                    & data[col.CIRCUMCISED]]) == 0
+
+    # no uncircumcised people have circumcision dates
+    assert len(data[~data[col.CIRCUMCISED]
+                    & data[col.CIRCUMCISION_DATE].notnull()]) == 0
+
+    # all circumcised people have circumcision dates
+    assert len(data[data[col.CIRCUMCISED]]) == \
+           len(data[data[col.CIRCUMCISED] & data[col.CIRCUMCISION_DATE].notnull()])
```
mmcleod89, line 160:
```suggestion
    assert (data[col.CIRCUMCISED] == data[col.CIRCUMCISION_DATE].notnull()).all()
```
Current test checks that circumcised => circumcision date but doesn't check that some people might have a circumcision date but not be circumcised. It is I think simpler to check that the boolean array for circumcision is equivalent to circumcision date being not null, and this will check for the implication in both directions.

mmcleod89:
```suggestion
assert (data[col.CIRCUMCISED] == data[col.CIRCUMCISION_DATE].notnull()).all()
```
We can combine these tests into a single statement by just checking that these two boolean arrays are identical. However this could make it slightly harder to see how things are going wrong if it fails.

src/tests/test_circumcision.py
mmcleod89: Same is also true for some other `len` statements throughout where you have boolean arrays. Which do you find more intuitive?

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+    pop.circumcision.covid_disrup_affected = 0
+    pop.circumcision.vmmc_disrup_covid = 0
+    pop.circumcision.mc_int = 2008
+    pop.circumcision.year_interv = 2022
+
+    # evolve population
+    pop.data.age += time_step.days / 365
+    pop.circumcision.update_vmmc(pop)
+    # check no VMMC occurs until after mc_int
+    assert len(pop.data[pop.data[col.VMMC]]) == 0
```
mmcleod89: The date is 1/12/2008, and the introduction of VMMC is 2008, so shouldn't VMMC already be active? 
I one update enough to test this? i.e. do we have enough boys that we expect at least one to get circumcised in this timestep? It might be better to check multiple updates before the VMMC introduction date.

src/hivpy/circumcision.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Update birth circumcision for newly born males.
+        COVID disruption is factored in.
+
+        Requires birth circumcision to be initialised with
+        `init_birth_circumcision_born` instead of
+        `init_birth_circumcision_all` in order to work as expected.
+        """
+        # covid disruption causes circumcision probability to be 0
+        if (self.covid_disrup_affected != 1) & (self.vmmc_disrup_covid != 1):
```
mmcleod89: `covid_disrup_affected` and `vmmc_disrup_covid` should probably be booleans since they can only be 1 or 0, and it will also simplify `if` statements

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+                           & (pop.data[col.AGE] < 20)])
+    male_population = pop.data[(pop.data[col.SEX] == SexType.Male)
+                               & (pop.data[col.AGE] >= 15)
+                               & (pop.data[col.AGE] < 20)]
+    prob_circ = ((2013 - pop.circumcision.mc_int) + (2019 - 2013)
+                 * pop.circumcision.rel_incr_circ_post_2013
+                 * pop.circumcision.circ_inc_15_19) * pop.circumcision.circ_inc_rate
+    # cap probability at 1
+    if (prob_circ > 1):
+        prob_circ = 1
```
mmcleod89: can also use `prob_circ = min(prob_circ, 1)`

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+    pop.circumcision.mc_int = 2008
+    pop.circumcision.year_interv = 2022
+
+    # evolve population
+    pop.data.age += time_step.days / 365
+    pop.circumcision.update_vmmc(pop)
+    circ_males = pop.data.index[(pop.data[col.SEX] == SexType.Male) & pop.data[col.CIRCUMCISED]]
+    pop.date += time_step
+    pop.data.age += time_step.days / 365
+    pop.circumcision.update_vmmc(pop)
```
mmcleod89, line 301: In the SAS code I think that this also affects birth circumcision, although I'm not sure if that is intentional. We should raise this with the HIV team as this would impact our birth circumcision code

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+    prob_circ = ((2013 - pop.circumcision.mc_int) + (2019 - 2013)
+                 * pop.circumcision.rel_incr_circ_post_2013
+                 * pop.circumcision.circ_inc_15_19) * pop.circumcision.circ_inc_rate
+    # cap probability at 1
+    if (prob_circ > 1):
+        prob_circ = 1
+    mean = len(male_population) * prob_circ
+    stdev = sqrt(mean * (1 - prob_circ))
+    # check circumcised value is within 3 standard deviations
+    assert mean - 3 * stdev <= no_vmmc <= mean + 3 * stdev
```
mmcleod89: Is there a reason not to use `general_circumcision_checks` here and for other vmmc tests?

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,435 @@
...
+        # check circumcisied people remain circumcised each step
+        new_circ_males = pop.data.index[(pop.data[col.SEX] == SexType.Male)
+                                        & pop.data[col.CIRCUMCISED]]
+        assert circ_males.isin(new_circ_males).all()
+
+    # nobody under 10 has been circumcised
+    assert len(pop.data[(pop.data[col.AGE] < 10) & (pop.data[col.VMMC])]) == 0
+
+
+def test_vmmc_case_i1():
```
mmcleod89: Would be good if the VMMC tests could be parameterised for different age groups, as we are checking the 15-20 age groups but not 20-30 or 30+

src/hivpy/circumcision_data.py
```diff
@@ -0,0 +1,44 @@
...
+from .common import DiscreteChoice
+
+
+class CircumcisionData:
+    """
+    Class to hold and interpret circumcision data loaded from the yaml file.
+    """
+
+    # TODO: This is ripped directly from sex_behaviour_data.py,
+    # we should make a new data reader module to store functions like this.
```
mmcleod89, line 15: Agreed! We need to remember to make an issue for this. It might make sense to have a data class, which all the specific data classes inherit from, that has these methods

src/hivpy/circumcision.py
```diff
@@ -0,0 +1,182 @@
...
+        circ_born_population = population.index[population[col.CIRCUMCISED]
+                                                & (population[col.AGE] >= 0.25)]
+        circ_unborn_population = population.index[population[col.CIRCUMCISED]
+                                                  & (population[col.AGE] < 0.25)]
+        # use current simulation start date as circumcision date for born individuals
+        population.loc[circ_born_population, col.CIRCUMCISION_DATE] = date
+        # find date where each unborn individual's age would be 0.25
+        population.loc[circ_unborn_population,
+                       col.CIRCUMCISION_DATE] = population[col.AGE].transform(
+                                                lambda x: date - timedelta(days=(x+0.25)*365))
```
mmcleod89: I think this should be `(x-0.25)` because `x` is their (currently negative) age, so the time until they get circumcised is $(-x) + 0.25$ or $- (x-0.25)$

src/hivpy/circumcision.py
```diff
@@ -0,0 +1,182 @@
...
+        # the function passed to transform_group
+        self.date = pop.date  # should year be float instead of int?
+
+        # no further circumcision
+        if (self.vmmc_disrup_covid == 1) \
+           | ((self.year_interv <= self.date.year)
+              & (self.circ_inc_rate_year_i == 2)) \
+           | ((self.year_interv + 5 <= self.date.year)
+              & (self.circ_inc_rate_year_i == 4)):
+            self.test_link_circ_prob = 0
```
mmcleod89:
```suggestion
        if (self.vmmc_disrup_covid == 1) \
           | ((self.year_interv <= self.date.year) & (self.circ_inc_rate_year_i == 2)) \
           | ((self.year_interv + 5 <= self.date.year) & (self.circ_inc_rate_year_i == 4)):
            self.test_link_circ_prob = 0
```
I personally find this formatting a bit easier to read as it requires less parsing of brackets between lines, but up to you which you prefer

src/hivpy/circumcision.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # only begin VMMC after a specific year
+        elif self.mc_int < self.date.year:
```
mmcleod89: SAS code treats the calendar date as a float so 2008.25 for example is march 2008. So we should activate VMMC the beginning of `mc_int`, whereas this will wait for the start of the next year

src/hivpy/circumcision.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        age_mod = 1
+        # age group 1 has no modifier in most cases
+        if age_group == 2:
+            age_mod = self.circ_red_20_30
+        elif age_group == 3:
+            age_mod = self.circ_red_30_50
+
+        calc_date = self.date.year
+        # cap date at 2019 for calculations
```
mmcleod89: Do we want 2019 and 2013 to be hardcoded?

src/hivpy/circumcision.py
```diff
@@ -0,0 +1,182 @@
...
+        age group and returns VMMC outcomes.
+        """
+        age_mod = 1
+        # age group 1 has no modifier in most cases
+        if age_group == 2:
+            age_mod = self.circ_red_20_30
+        elif age_group == 3:
+            age_mod = self.circ_red_30_50
+
+        calc_date = self.date.year
```
mmcleod89, line 171: We don't need to do this with integers, we can do it continuously as in `update_rred_population`. I don't expect it to make a noticeable difference though.

tutorials/circumcision.md
```diff
@@ -0,0 +1,63 @@
...
+- `src/hivpy/data/circumcision.yaml` - Circumcision data and variables.
+- `src/hivpy/circumcision_data.py` - A class for storing data loaded from `circumcision.yaml`.
+
+If there are any circumcision-related variables you would like to change before running your simulation, please change them in `circumcision.yaml`.
+
+### Population Initialisation
+
+Birth circumcision is initialised when population data is first created. There are currently two methods for initialising birth circumcision:
+
+- `init_birth_circumcision_all` - Initialises circumcision at birth for the entire male population, both born and unborn. Circumcision dates for all born individuals are recorded as the date of the start of the simulation, but cirucmcision dates for unborn individuals are calculated by finding the dates at which each individual's age would be 0.25.
```
mmcleod89, line 16: We should mention here that this doesn't currently take into account COVID interruption

mmcleod89 (commented):
Module works well, most comments are stylistic so depends on how you feel. A few minor tweaks needed for consistency like with the onset of vmmc dates. 

The tutorial looks really good, I like the clear instructions on how to modify the code if they want to switch between methods.

src/tests/test_circumcision.py
pineapple-cat: In my opinion `sum` statements look better because they don't require as much repetition of `pop.data`, so I'd be happy to use them instead of `len`.

src/tests/test_circumcision.py
pineapple-cat: That's why I had them as separate statements in the first place, but considering we should be past the most painful bugs with circumcision outcomes and dates not lining up I think it would be fine to merge them into one statement now.

src/tests/test_circumcision.py
pineapple-cat: The only reason I didn't was because I expect `test_vmmc_basic` (acting as my `test_vmmc_case_i0`) to catch if anything untoward happens in the general checks. The population is entirely male so there's no reason to check if any females have been marked as circumcised, and I expect that if not all circumcised people have circumcision dates (or if some uncircumcised people have circumcision dates) this issue will have been caught earlier. I could use `general_circumcision_check`, I just think most of the checks are redundant.

src/hivpy/circumcision.py
pineapple-cat: I don't think it matters for 2013 since that value will always stay where it is unless the way the model works is changed (e.g. if `rel_incr_circ_post_2013` is no longer relevant), but I could give both of these years variable names and put them in the yaml file for separation's sake.

src/hivpy/circumcision.py
pineapple-cat: I agree that this formatting is easier to read, but flake8 will complain about the lines being too long (> 100 characters).

src/hivpy/circumcision.py
pineapple-cat: In that case changing this to `self.vmmc_start_date <= self.date.year` should give us the behaviour we want and let us keep the year as an int.

src/hivpy/circumcision.py
mmcleod89: I think we should probably consider making the allowable line length a little longer (maybe 120 characters) given that a) everyone's IDE can display more than 100 characters these days and b) we have a lot of long variables names which makes keeping things under the 100 character limit difficult. This is a common problem and had led to a lot of weird formatting in some of the modules

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,345 @@
...
+        # increases in 15-19 year olds after policy_intervention_year
+        pop.circumcision.circ_policy_scenario = 1
+        set_covid(pop.circumcision, False)
+        set_vmmc_default_dates(pop.circumcision)
+
+        # evolve population
+        pop.data.age += time_step.days / 365
+        pop.circumcision.update_vmmc(pop)
+        # nobody under 15 has been circumcised
+        assert sum((pop.data[col.AGE] < 15) & (pop.data[col.VMMC])) == 0
```
mmcleod89: This test is redundant here because we've set everyone's age to 18, 25, or 40

src/tests/test_circumcision.py
```diff
@@ -0,0 +1,345 @@
...
+                                            & pop.data[col.CIRCUMCISED]]
+            assert circ_males.isin(new_circ_males).all()
+
+        # nobody under 10 has been circumcised
+        assert sum((pop.data[col.AGE] < 10) & (pop.data[col.VMMC])) == 0
+
+
+def test_vmmc_case_1():
+
+    test_ages = [18, 25, 40]
```
mmcleod89: Should this include younger ages so we can see the effect on 10-14 year olds?

pineapple-cat: Seems reasonable, I can add 14 as a test age where we expect to see no VMMC after the policy intervention year.

mmcleod89 (approved):
Super! I'm happy with the new test changes and the documentation is looking absolutely fab :)

PR #93, opened 2022-10-25
Add columns for previous timesteps; fix some sex behaviour bugs

mmcleod89:


src/hivpy/population.py
```diff
@@ -66,6 +75,62 @@ def _create_population_data(self):
...
+            self.data[(col.name, i)] = init_val + 1
+
+    def get_sub_pop(self, conditions):
+        index = reduce(operator.and_,
+                       (op(self.data[self.get_correct_column(var)], val) for (var, op, val) in conditions))
+        return index
+
+    def apply_vector_func(self, params, func):
+        param_cols = list(map(self.get_correct_column, params))
+        return func(*param_cols)
```
ageorgou: You can also use the unpacking (`*`) directly with any iterable, without converting it to a list:
```suggestion
        param_cols = map(self.get_correct_column, params)
        return func(*param_cols)
```

src/hivpy/population.py
```diff
@@ -66,6 +75,62 @@ def _create_population_data(self):
...
+            return self.data.loc[sub_pop, var_col]
+
+    def set_present_variable(self, target: str, value, sub_pop=None):
+        present_col = self.get_correct_column((target, 0))
+        if sub_pop is None:
+            self.data[present_col] = value
+        else:
+            self.data.loc[sub_pop, present_col] = value
+
+    def get_correct_column(self, param_info):
```
ageorgou: I would mark this as "private" by naming it `_get_correct_column`, even though it doesn't actually restrict the visibility. Same for `make_column_tuple`.

src/hivpy/population.py
```diff
@@ -66,6 +75,62 @@ def _create_population_data(self):
...
+    def get_sub_pop(self, conditions):
+        index = reduce(operator.and_,
+                       (op(self.data[self.get_correct_column(var)], val) for (var, op, val) in conditions))
+        return index
+
+    def apply_vector_func(self, params, func):
+        param_cols = list(map(self.get_correct_column, params))
+        return func(*param_cols)
+
+    def get_variable(self, var, sub_pop=None):
```
ageorgou: I think that having `dt` as a separate, explicit optional argument here may be clearer, rather than (sometimes) packaging it into `var`.  Then you wouldn't need `make_column_tuple` either? (with corresponding small changes in `get_correct_column`)

src/hivpy/population.py
```diff
@@ -66,6 +75,62 @@ def _create_population_data(self):
...
+        else:
+            self.data.loc[sub_pop, present_col] = value
+
+    def get_correct_column(self, param_info):
+        (param, dt) = self.make_column_tuple(param_info)
+        col_index = (self.step + dt) % self.variable_history[param]
+        return (param, col_index)
+
+    def set_variable_by_group(self, target, groups, func, use_size=True, sub_pop=None):
+        """Sets the value of a population variable at the present time step by calling transform group."""
```
ageorgou: Yes, this seems like a very handy interface function for a commonly-used operation.

src/hivpy/population.py
```diff
@@ -78,20 +143,23 @@ def transform_group(self, param_list, func, use_size=True, sub_pop=None):
...
-        # HIV_STATUS is just a dummy column to allow us to use the transform method
+        # Use Dummy column to in order to enable transform method and avoid any risks to data
         def general_func(g):
             if len(param_list) == 1:
                 args = [g.name]
             else:
                 args = list(g.name)
+
+            g = list(map(lambda x: self.get_correct_column(x), g))
+
```
ageorgou: Not sure what this is doing, but it looks like it will not work with the following lines (`g` will be a list and will not longer have a `size` attribute).

Should the second argument to `map` be `args` instead of `g`?

ageorgou: Also, you don't need the lambda here. You could simply `map(self.get_correct_column, ...)`

src/hivpy/population.py
```diff
@@ -66,6 +75,62 @@ def _create_population_data(self):
...
+        return func(*param_cols)
+
+    def get_variable(self, var, sub_pop=None):
+        var_col = self.get_correct_column(var)
+        if sub_pop is None:
+            return self.data[var_col]
+        else:
+            return self.data.loc[sub_pop, var_col]
+
+    def set_present_variable(self, target: str, value, sub_pop=None):
```
ageorgou, line 157: If this turns out to be too long a name, I think `set_variable` is clear enough, as no one should be setting past values anyway!

src/hivpy/population.py
```diff
@@ -66,6 +75,62 @@ def _create_population_data(self):
...
+        """Sets the value of a population variable at the present time step by calling transform group."""
+        target_col = self.get_correct_column((target, 0))
+        if sub_pop is None:
+            self.data[target_col] = self.transform_group(groups, func, use_size)
+        else:
+            self.data.loc[sub_pop, target_col] = self.transform_group(groups, func, use_size, sub_pop)
+
+    def make_column_tuple(param):
+        if (type(param) == str):
+            return (param, 1)
```
ageorgou: Should the default value be 0 or 1 (as in the setting methods)?

ageorgou (commented):
I've not gone through all the updates for assigning variables in detail, but this generally makes sense.

Some general comments:
- I agree that the indirection for accessing columns of the dataframe is a good idea.
- Another thing to consider could be defining the special `__getitem__` function (to essentially call `get_variable` under the hood), so that you can access columns as e.g. `my_population["age"]` or `my_population["age", subpopulation]`.
- Similar for setting variables with `__setitem__` although that's perhaps more complicated because you may want to distinguish between subpopulations and groups. And going through the code now, I realise you'd have the same issue with the `get_variable`, because you want it to also take time lags. So maybe this is not such a good suggestion after all.
- I wonder whether the variable history is more naturally defined externally rather than on the Population itself (perhaps with the column names?) Although this would be more for the sake of organisation, I don't think it would offer anything more - and could be done later if decided anyway.
- (if you haven't seen already!) To avoid the circular imports due to the type hint, you can use a constant that's provided by the `typing` module; see [a post with more details](https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/) and [this StackOverflow question](https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports). Unfortunately, it's a bit clunky. The other option is to remove the type hint.

mmcleod89:
> (if you haven't seen already!) To avoid the circular imports due to the type hint, you can use a constant that's provided by the typing module; see [a post with more details](https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/) and [this StackOverflow question](https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports). Unfortunately, it's a bit clunky. The other option is to remove the type hint.

I hadn't actually seen that before; I spent quite a bit of time yesterday trying to figure out how to get around the circular imports and just ended up having to take them out on my working branch! The trouble with that is that it makes development a lot harder because  a) you don't know off hand what type something is _meant_ to be, and b) you don't get any of the additional information like what functions are available and function signatures. I did consider restructuring so that the other modules point to the population rather than the population pointing to the modules, so the evolve step wouldn't be called on population but would be a _simulation_ method instead, and the simulation would tell each of the modules to do their thing and each module would act on the population. This would solve the circular import issue and still makes conceptual sense, as one might readily think of the modules as being aspects of the simulation rather than aspects of the population itself.

src/hivpy/population.py
mmcleod89: Yeah there is basically just the issue of clarity vs verbosity; I had the make_column_tuple because very few things need t-2 or t-3 so it saves some space to not have to type it every time, although at the expense of clearer logic. The other thing being that having a tuple of (variable, dt) throughout helps a bit with consistency, since we have methods like the groupby wrapper which take a list of variables as those tuples, which is clearer than providing separate arguments for a list of variables and a list of time difference. 
It might be clearer if we use some kind of struct instead of a tuple so that we can specify the type more explicitly?

src/hivpy/population.py
mmcleod89: This was mostly just wrong and has been moved somewhere else; I think I was tired when I wrote this! This map applies to the list of params, to get the columns for the correct time periods.

src/hivpy/population.py
mmcleod89: I chose the default of 1 because usually we are getting columns from t-1 in order to calculate new values, but I admit it's not super clear so maybe we shouldn't have a default at all.

mmcleod89:
New methods for dealing with dataframe are working and are backwards compatible so work with existing code (as well as new code written for e.g. circumcision). I have not gone back through and changed all instances of previous code style as it may be a lot to do in one go and I'd like to move forward, but we can migrate to this way of doing things as we go ahead which will make it easier to change how we handle the dataframe in the future (especially in the case that Pandas makes changes to the way that it works!!)

src/hivpy/demographics.py
```diff
@@ -217,16 +224,17 @@ def _probability_of_death(self, sex: SexType, age_group: int) -> float:
         prob_of_death = 1 - exp(-rate / 4)
         return prob_of_death
 
-    def determine_deaths(self, pop) -> pd.Series:
+    def determine_deaths(self, pop: Population) -> pd.Series:
```
pineapple-cat, line 227: Potential style note for later: in various modules we use `pop` and `population` as variable names both for `Population` and `pd.DataFrame` without any particular convention. For consistency, we could stick to one (either `pop` or `population`), or we could use one variable name for `Population` and the other for `pd.DataFrame` so it becomes easier to intuit the type a given function is meant to be working with.

src/hivpy/sexual_behaviour.py
```diff
@@ -162,100 +175,110 @@ def _assign_new_sex_group(self, sex, group, rred, size):
...
-            indices = selector(pop_data, art_adherence=(operator.lt, self.adherence_threshold))
-            pop_data.loc[indices, col.RRED_ART_ADHERENCE] = self.rred_art_adherence
-
-    def update_rred_age(self, pop_data):
-        over_15s = selector(pop_data, age=(operator.ge, 15))
-        age = pop_data.loc[over_15s, col.AGE]
-        sex = pop_data.loc[over_15s, col.SEX]
+            self.update_rred_art_adherence(population)
+
+        def comined_rred(age, adc, balance, diagnosis, personal, ltp, art):
```
pineapple-cat: I take it this should be `combined_rred`?

pineapple-cat (approved):
Would be great to see a tutorial at some point, but otherwise looks ready to go.

src/hivpy/demographics.py
mmcleod89: This is a good point, I think perhaps I'd lean towards `pop` for brevity given that long line lengths are a perennial problem in this code, as long as that doesn't look confusing. Ideally I'd like us to use the dataframe directly as little as possible after these changes have been merged, and the dataframe shouldn't ever be referred to as a "population" as I find that confusing given that we have a different class called `Population`. If the dataframe is for any reason referred to directly, it should be called something like `pop_data`.

src/hivpy/sexual_behaviour.py
mmcleod89: Yes it should!

PR #91, opened 2022-10-14
Improve Sphinx documentation

pineapple-cat:
Moved documentation location to a more intuitive place:
1. Created an empty `docs/` folder at the same level as `src/` and then moved to that folder and ran `sphinx-quickbuild`.
2. Adjusted `conf.py` to include the `sphinx.ext.autodoc` extension and ran `sphinx-apidoc -o . ../src/hivpy` to set output and module paths.
3. Ran `make html` to get html docs in `_build/html`.

Still looking into how to change documentation formatting.

Issue #33, opened 2022-02-03
Track variables while simulation is happening


ageorgou:
As reported in the demo (#25):

It would be useful to be able to see a subset of the variables (e.g. HIV status) for the whole or a subset of the population, before the whole simulation is complete. This is so we can quickly check for any unexpected behaviour without having to wait for too long, and at a higher granularity (at the level of individuals) compared to the "normal" simulation outputs.

One solution could be via logging. It may be enough to add debug-level logging calls at a reasonable place, along with some sample code for selecting the variables and individual characteristics of interest. Or it may be better controlled by the config. An alternative would be to use the debugger.

These files could end up very large very quickly, so it would be good to avoid storing them long-term if possible. At the very least, we could use a naming scheme that makes them easy to identify and delete, or we could try to store them in a temporary location? (deleting them automatically at the end of the simulation may be too risky)

ageorgou:
More details from discussion today:
- Being able to compare across time is important, so a log file would probably be more useful than the debugger.
- Does not have to happen for the whole simulation, but only specific time steps (e.g. every time step between particular dates).
- As above, not interested in all the variables nor the whole population. Common selection criteria are sex and age, but others are also used (could be anything in principle). Sometimes we select just so that we have a more manageable subset.

PR #96, opened 2022-11-01
Start richer logging

ageorgou:
Addresses #33.

So far:
- Adds a description of how the new feature will work, to be included as a tutorial
- Refactors the existing logging to:
    - allow multiple sub-loggers (as we did in tests and demographics)
    - keep the functionality of "writing" of the population variables of interest as a separate class

Still needed:
- [ ] actually parsing the configuration and determining what should be written at each step!
- [ ] writing of the population at each step
- [ ] moving the tutorial to the docs directory (dependent on #91) and adding it in the table of contents so it's part of the generated documentation

### DIFF
--- a/logging_tutorial.rst
+++ b/logging_tutorial.rst
@@ -1,0 +1,99 @@
+Logging tutorial
+================
+
+While running the model, it is sometimes useful to monitor the progress
+of certain characteristics across time. This can be used, for instance,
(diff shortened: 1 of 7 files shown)


pineapple-cat (approved):
Tutorial looks good and examples are clear, I'll finish up with moving the updated Sphinx docs and then when everything is merged I'll put the tutorial in the right directory.

PR #99, opened 2022-11-15
Diagonalisable sex behaviour matrices check

mmcleod89:
Checks that the (transpose of) transition matrices for sexual behaviour groups are diagonalisable, and thus can be used in order to calculate new matrices for arbitrary time-steps


mmcleod89:
Currently this is not needed as variable timestep is unlikely to be implemented (number of new partners has no self consistent transformation)

PR #102, opened 2022-12-02
Circumcision variable name suggestions

mmcleod89:
Some name changes; you can adopt them if you like. You'll understand the module better than I do I expect so if you think any of my names are misleading let me know. 

I would like better names for `year_interv` and `test_link_circ_prob` but I don't really understand what they are yet. 

I'm a little uneasy with using the `15_19`, `20_30` naming conventions; I usually prefer to keep variations of the same variable in a vector. (You could then use `circ_rate_change[age_group]` instead of `if` statements to assign your variable in `calc_prob_circ`. I've left it as-is for now, but something to consider as we don't want to end up with their style of using variable names instead of indexing / mapping.

### DIFF
--- a/src/hivpy/circumcision.py
+++ b/src/hivpy/circumcision.py
@@ -15,18 +15,18 @@ class CircumcisionModule:
         # init cirumcision data
         with importlib.resources.path("hivpy.data", "circumcision.yaml") as data_path:
             self.c_data = CircumcisionData(data_path)
-        self.mc_int = self.c_data.mc_int
+        self.vmmc_start_date = self.c_data.vmmc_start_date
         self.year_interv = self.c_data.year_interv
         self.test_link_circ_prob = self.c_data.test_link_circ_prob
-        self.circ_inc_rate_year_i = self.c_data.circ_inc_rate_year_i
+        self.circumcision_increase_scenario = self.c_data.circumcision_increase_scenario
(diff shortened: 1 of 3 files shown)


pineapple-cat (approved):
Looks reasonable to me, though I think I'll change `circumcision_increase_scenario` to just `circ_scenario` since most of the scenarios don't even include an increase to circumcision. I also don't like the age group naming conventions and I'll try your suggestion of keeping them in a vector in the main module code, but I'll probably rename them to have more consistent numbers in the data class and the yaml (i.e. the youngest age group will either become `15-20` or the other two will become `20-29` and `30-49`).

The variable `year_interv` is linked to circumcision scenarios and several other various bits, such as changes in ART coverage. I'm also not entirely sure what intervention this year is meant to represent, so it would be worth asking for more information.

Additinally, `test_link_circ` and `test_link_circ_prob` are only used here:
```
if test_link_circ=1 and t ge 2 and tested_tm1=1 and registd_tm1 ne 1 then do;
    uc=rand('uniform'); if uc < test_link_circ_prob  and vmmc_disrup_covid ne 1 then do; mcirc=1;new_mcirc=1; end;
end;
```
They link circumcision to HIV testing, but I'm not sure what name we could call them that is both relatively descriptive and concise. Maybe `circ_after_test` and `prob_circ_after_test`?

PR #104, opened 2022-12-16
Hard to reach individuals

### LINKED ISSUE #84: Hard to reach populations
- Implement hard to reach people (as a boolean column?) 
- Hard to reach should be applied to circumcision module as well

pineapple-cat:
Added hard to reach column that currently affects circumcision. HIV testing and PrEP is not modelled yet.

### DIFF
--- a/src/hivpy/demographics.py
+++ b/src/hivpy/demographics.py
@@ -198,7 +198,7 @@ class DemographicsModule:
 
     def initialise_hard_reach(self, population):
         # base probabilities for being hard to reach
-        self.prob_hard_reach_f = round(rng.uniform(0.05, 0.15), 2)
+        self.prob_hard_reach_f = round(rng.uniform(0.05, 0.2), 2)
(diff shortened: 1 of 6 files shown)


src/hivpy/demographics.py
```diff
@@ -196,6 +196,21 @@ def initialise_age(self, count):
 
         return age_distribution.gen_ages(count)
 
+    def initialise_hard_reach(self, population):
+        # base probabilities for being hard to reach
+        self.prob_hard_reach_f = round(rng.uniform(0.05, 0.15), 2)
```
mmcleod89:
```suggestion
        self.prob_hard_reach_f = round(rng.uniform(0.05, 0.2), 2)
```
I think upper limit should be 0.02? Based on 
```
p_hard_reach_w=0.05+(rand('uniform')*0.15) 
```
so adding 0.05 to upper and lower limits.

mmcleod89 (commented):
Looks good to me! The covid flag for initialisation is a good addition too. 
The only issue I think is checking the limits for the random number for hard to reach in women, otherwise I think it's ready to go

PR #105, opened 2022-12-19
Add pregnancy feature

### LINKED ISSUE #79: Pregnancy

pineapple-cat:
To implement:

- [x] Initialise fertility + number of children
- [x] Explicit pregnancy
- [x] Min wait after childbirth before next pregnancy
- [x] Different STP vs LTP pregnancy probability
- [x] Childbirth
- [x] Wanting no more children + max child cap
- [x] Antenatal care attendance
- [x] Prevention of mother to child transmission
- [x] Birth with infected child + infected child count

### DIFF
--- a/src/hivpy/pregnancy.py
+++ b/src/hivpy/pregnancy.py
@@ -19,13 +19,13 @@ class PregnancyModule:
             self.p_data = PregnancyData(data_path)
 
         self.can_be_pregnant = self.p_data.can_be_pregnant
+        self.rate_want_no_children = self.p_data.rate_want_no_children
         self.fold_preg = self.p_data.fold_preg
         self.inc_cat = self.p_data.inc_cat.sample()
-        self.rate_birth_with_infected_child = self.p_data.rate_birth_with_infected_child.sample()
         self.prob_pregnancy_base = self.generate_prob_pregnancy_base()  # dependent on time step length
+        self.rate_birth_with_infected_child = self.p_data.rate_birth_with_infected_child.sample()
+        self.max_children = self.p_data.max_children
         self.init_num_children_distributions = self.p_data.init_num_children_distributions
-        self.rate_want_no_children = 0.005
-        self.max_children = 10
 
     def generate_prob_pregnancy_base(self):
         """
@@ -54,8 +54,7 @@ class PregnancyModule:
         Initialise the number of children each female individual
         at or above the age of 15 starts out with.
         """
-        # TODO: ask about this
-        # negation of low fertility condition not present originally but seems logical to include
+        # get fertile female population above age 14
         female_population = pop.data.index[(pop.data[col.SEX] == SexType.Female)
                                            & (~pop.data[col.LOW_FERTILITY])
                                            & (pop.data[col.AGE] >= 15)]
@@ -87,8 +86,7 @@ class PregnancyModule:
         # TODO: this is needed for stp prob preg reduction but should
         # probably be moved elsewhere so it's only initialised once
         self.fold_tr_newp = pop.hiv_status.fold_tr_newp
-        # get sexually active female population
-        # to check for new pregnancies
+        # get sexually active female population to check for new pregnancies
         active_female_population = pop.data[(pop.data[col.SEX] == SexType.Female)
                                             & (pop.data[col.AGE] >= 15)
                                             & (pop.data[col.AGE] < 55)
@@ -122,7 +120,8 @@ class PregnancyModule:
             # TODO: change age group col name to be more descriptive
             pop.data.loc[active_female_population[pregnancy_ready].index, col.AGE_GROUP] = age_groups
             # calculate pregnancy outcomes
-            pregnancy = pop.transform_group([col.AGE_GROUP, col.WANT_NO_CHILDREN, col.NUM_PARTNERS],
+            pregnancy = pop.transform_group([col.AGE_GROUP, col.LONG_TERM_PARTNER,
+                                             col.NUM_PARTNERS, col.WANT_NO_CHILDREN],
                                             self.calc_preg_outcomes,
                                             sub_pop=active_female_population[pregnancy_ready].index)
             # assign outcomes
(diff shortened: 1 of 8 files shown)


pineapple-cat:
Most basic pregnancy-related functionality has been implemented, with the exception of testing-related features (to be addressed in #109) and a few other issues mentioned in FIXME comments. An issue has also been set up for a pregnancy tutorial #114 to be completed later.

src/hivpy/pregnancy.py
```diff
@@ -0,0 +1,268 @@
...
+        # init pregnancy data
+        with importlib.resources.path("hivpy.data", "pregnancy.yaml") as data_path:
+            self.p_data = PregnancyData(data_path)
+
+        self.can_be_pregnant = self.p_data.can_be_pregnant
+        self.rate_want_no_children = self.p_data.rate_want_no_children  # dependent on time step length
+        self.date_pmtct = self.p_data.date_pmtct
+        self.pmtct_inc_rate = self.p_data.pmtct_inc_rate
+        self.fold_preg = self.p_data.fold_preg
+        self.inc_cat = self.p_data.inc_cat.sample()
```
mmcleod89, line 33: I think this highlights an issue with the code at the moment. `inc_cat` already exists in `DemographicsModule.params` so we need to find a way to expose the data modules, or the modules themselves, to one another. At the moment the only way to do so is through the population object, which would mean passing the population or the demographics module into the constructor of this module.

src/hivpy/pregnancy.py
```diff
@@ -0,0 +1,268 @@
...
+        if len(active_female_population) > 0:
+
+            # at least 6 months must pass since last childbirth
+            # (15 months since last pregnancy start date)
+            previously_pregnant = active_female_population[active_female_population[col.LAST_PREGNANCY_DATE].notnull()]
+            pregnancy_ready = ~active_female_population[col.LOW_FERTILITY]
+            if len(previously_pregnant) > 0:
+                # mask
+                pregnancy_ready = pregnancy_ready & (previously_pregnant[col.LAST_PREGNANCY_DATE]
+                                                     + timedelta(days=450) <= pop.date)
```
mmcleod89: I think this code may be incorrect. We have that someone is pregnancy ready if they are:

1. sexually active
2. not low fertility (I think this might be redundant because it is already incorporated into the `active_female_population`)
3. not recently pregnant 

I think the issue is that `previously_pregnant` is only true if it is not null _and_ the date was more than 6+9 months ago. But as far as I can tell this leaves out women who have never been pregnant and whose last pregnancy date is `None`. As far as I understand it, `pregnancy_ready` should be true if last pregnancy date is `None` _or_ the date is more than 15 months ago.

pineapple-cat: You're right, `~active_female_population[col.LOW_FERTILITY]` is redundant, I just needed it there to create an all True boolean series of the same length as `active_female_population`. Perhaps it would be less confusing to replace it with something like `[True] * len(active_female_population)`?

Also, I see what you're saying, but I don't think making the `pregnancy_ready` _and_ statement into an _or_ statement works because it would be the same as not checking whether a woman has been recently pregnant. The point is that the entire `active_female_population` should be capable of getting pregnant (which includes women who have never been pregnant and thus have a `None` last pregnancy date, otherwise my tests would be failing!), _unless_ they are currently in their "cooldown period" from being recently pregnant (i.e. `(previously_pregnant[col.LAST_PREGNANCY_DATE] + timedelta(days=450) <= pop.date)` is False). So for a woman who has _not_ been recently pregnant we have `pregnancy_ready = True & True = True`, but for a woman who _has_ been recently pregnant we have `pregnancy_ready = True & False = False`.

src/hivpy/pregnancy.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        # get pregnant population
+        pregnant_population = pop.data[pop.data[col.PREGNANT]]
+        # update probability of antenatal care attendance
+        self.prob_anc = min(max(self.prob_anc, 0.1) + self.rate_testanc_inc, 0.975)
+
+        # anc outcomes
+        r = rng.uniform(size=len(pregnant_population))
+        anc = r < self.prob_anc
+        pop.data.loc[pregnant_population.index, col.ANC] = anc
```
mmcleod89: This ANC state is persistant even after pregnancy is ended. Is this correct? Should ANC in one pregnancy imply ANC for all future pregnancies?

src/tests/test_pregnancy.py
```diff
@@ -0,0 +1,385 @@
...
+    # check that nobody is pregnant anymore
+    assert ~pop.data[col.PREGNANT].all()
+    # check that everyone now has one child
+    assert (pop.data[col.NUM_CHILDREN] == 1).all()
+
+    # test the pregnancy pause period
+    pop.pregnancy.update_pregnancy(pop)
+    pop.date += time_step
+    # check that there are still no pregnancies
+    assert ~pop.data[col.PREGNANT].all()
```
mmcleod89: This tests that _not all_ people are pregnant. To test that no people are pregnant you need `(~pop.data[col.PREGNANT).all()` i.e. you need to invert before you call `all()`.

src/tests/test_pregnancy.py
```diff
@@ -0,0 +1,385 @@
...
+
+    pop.pregnancy.update_pregnancy(pop)
+    pop.date += time_step
+    # check that everyone is pregnant again
+    assert pop.data[col.PREGNANT].all()
```
mmcleod89, line 175: I am a bit confused about this, has enough time passed for people to become pregnant again?

src/hivpy/pregnancy.py
mmcleod89: Right, but the problem is what if we have a women who has `col.LAST_PREGNANCY_DATE` as `None`? Then we have `(previously_pregnant[col.LAST_PREGNANCY_DATE] + timedelta(days=450) <= pop.date)` is `False` and therefore `pregnancy_ready = True & False = False` don't we?

mmcleod89 (approved):

PR #117, opened 2023-02-28
Add HIV testing feature

### LINKED ISSUE #107: General Population HIV Testing (no diagnosis)
Basic implementation of marking people as getting testing in a given timestep for the general population. Start point for testing module
- In SAS code lines 4349-4387
- Two subgroups: first time tested and repeat testers 
- Can refactor setting testing variables into a function as this is repeated throughout SAS code 
- Dummy columns for e.g. newp_lasttest (number of short term condomless sex partners since last test) and np_lasttest (total number of condomless sex partners since last test i.e. including long term partner)
- eff_test_targeting is a personal variable and test_targeting is a population variable, but currently eff_test_targeting is just set to test_targeting.

pineapple-cat:
To implement:

- [x] Update tested status for first time and repeat testers
- [x] Add dummy columns for number of partners
- [x] Add `eff_test_targeting` and `test_targeting`
- [x] More thorough unit tests (e.g. make sure diagnosed people aren't being tested, check partner columns are being properly reset, test `calc_prob_test` logic)
- [x] Add `no_test_if_np0` to config file
- [x] Add `eff_max_freq_testing`
- [x] Add `incr_test_targeting_year_i` scenarios


src/hivpy/hiv_testing.py
```diff
@@ -0,0 +1,145 @@
...
+        self.testing_disrup_covid = False
+
+    def update_hiv_testing(self, pop):
+        """
+        Update which individuals in the population have been tested.
+        COVID disruption is factored in.
+        """
+        # testing occurs after a certain year if there is no covid disruption
+        if ((pop.date.year >= (self.date_start_testing + 5.5))
+           & ((not self.covid_disrup_affected) | (not self.testing_disrup_covid))):
```
mmcleod89: I think there needs to be a change here, but we should ask the HIV team about this. 

- In the SAS code `date_start_testing` is 2003.5, and so `date_start_testing + 5.5` is 2009 
- `pop.date.year` is an int, which is okay if we're comparing to 2009, but if this value were not a whole number then we would be in trouble. It woule be better to calculate this as an actual date and compare rather than using the same calculation as in the SAS code, because they represent their date as a decimal number like 2003.5 for the middle of 2003
- This calculation also just strikes me as suspicious; would we still be adding 5.5 if `date_start_testing` were not 2003.5? Is there a significance to 5.5 years? I think we should check up on this.

mmcleod89: In general we also have the potential issue of an int >= a float may fail on the equal case; should be fine in this case since exactly 0.5 is representable in FP but something we should avoid in general I think when the exact case is important.

mmcleod89: `date_start_testing + 5.5` also comes up several times in the following code so might be worth giving it its own variable. But again we can address that after we've found out a bit more about what it means

src/hivpy/hiv_testing.py
```diff
@@ -0,0 +1,145 @@
...
+                                                    timedelta(days=self.days_to_wait[self.eff_max_freq_testing])),
+                                                   (col.LAST_TEST_DATE, op.eq, None)]
+                                                  ])
+
+            # first time testers
+            untested_population = pop.apply_bool_mask(~pop.data.loc[testing_population, col.EVER_TESTED],
+                                                      testing_population)
+            # repeat testers
+            prev_tested_population = pop.apply_bool_mask(pop.data.loc[testing_population, col.EVER_TESTED],
+                                                         testing_population)
```
```suggestion
# first time testers
            untested_population = pop.apply_bool_mask(~pop.get_variable(col.EVER_TESTED, testing_population),
                                                      testing_population)
            # repeat testers
            prev_tested_population = pop.apply_bool_mask(pop.get_variable(col.EVER_TESTED, testing_population),
                                                         testing_population)
```
To avoid direct `.data` accesses

src/tests/test_hiv_testing.py
```diff
@@ -0,0 +1,190 @@
...
+        # evolve population
+        pop.hiv_testing.update_hiv_testing(pop)
+
+        # get people that were just tested
+        tested_population = pop.get_sub_pop([(col.LAST_TEST_DATE, op.eq, pop.date)])
+        # check that partner numbers have been reset
+        assert sum(pop.get_variable(col.NP_LAST_TEST, sub_pop=tested_population)) == 0
+        assert sum(pop.get_variable(col.NSTP_LAST_TEST, sub_pop=tested_population)) == 0
+        # check that no people just tested were already diagnosed with HIV
+        assert (~pop.get_variable(col.HIV_STATUS, sub_pop=tested_population)).all()
```
mmcleod89: This line feels like it would be more at home in a different test e.g. the first time & repeat testers checks, or in its own; I think `test_partner_reset_after_test` suggests a very specific test so if it fails on this line people might be confused about what actually is going wrong.

mmcleod89 (changes requested):
Generally looks good to me, just a couple of minor tweaks I'd like to address. I'll also create a new issue to make sure that we add NP_LAST_TEST etc. tracking to the sex behaviour logic!

mmcleod89:
You should also add a call to the hiv testing code to the population.evolve method, otherwise it won't execute in a simulation!

pineapple-cat:
Haven't done anything about the problems that may arise from `pop.date.year` being an int rather than a float (might be worth addressing as a separate issue across multiple modules) but everything else should be good to go now.

mmcleod89 (commented):
Okay that looks fine to me. Re the date issue, I think a good solution would be to have it be an actual datetime.date object so it's consistent with the way we store dates in the population. This could be something that we implement in the file-loading class though since that should do the job of converting inputs into useable formats, so we can create a load_date method or similar to change 2003.5 into June 2003.

mmcleod89 (approved):

PR #118, opened 2023-03-07
convert pregnancy code to new access functions

mmcleod89:


pineapple-cat (approved):
`sub_pop=None` doesn't need to be in `get_sub_pop` since it's not used, but aside from that I'm happy with and understand these changes.

PR #122, opened 2023-05-15
Add general data reader

### LINKED ISSUE #103: Create data reader module
The `SexualBehaviourData` and `CircumcisionData` classes currently have some repeated code (`_get_discrete_dist`). It would be useful to make a general data class for all other data classes to inherit from, as it is likely we will end up with more overlap down the line.

pineapple-cat:
Refactoring data readers for better code reuse.

### DIFF
--- a/src/hivpy/circumcision_data.py
+++ b/src/hivpy/circumcision_data.py
@@ -1,30 +1,16 @@
-import numpy as np
-import yaml
-
 from hivpy.exceptions import DataLoadException
 
-from .common import DiscreteChoice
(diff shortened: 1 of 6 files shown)


pineapple-cat:
Technically `DemographicsData` doesn't really use anything from `DataReader` and thus doesn't _need_ to inherit from it, but I decided to add the inheritance anyway for consistency (and in case it ever does need to make use of some `DataReader` functions in the future).

mmcleod89 (approved):

PR #125, opened 2023-05-17
Add ANC HIV testing

### LINKED ISSUE #109: Add ANC HIV testing to Pregnancy Module
To follow #79 #81 #107 

ANC testing in SAS code lines: 4210 - 4230. Change to additional HIV test 3 months after birth to be 2 tests during pregnancy. 

Should make use of the testing module to call function to set testing variables, but testing itself may be called from Pregnancy module if that makes more sense.

pineapple-cat:
Addresses #109.

### DIFF
--- a/src/hivpy/hiv_testing.py
+++ b/src/hivpy/hiv_testing.py
@@ -84,29 +84,70 @@ class HIVTestingModule:
                                              self.calc_testing_outcomes,
                                              sub_pop=untested_population)
                 # set outcomes
-                pop.set_present_variable(col.EVER_TESTED, tested, sub_pop=untested_population)
-                # set last test date
-                pop.set_present_variable(col.LAST_TEST_DATE, pop.date,
-                                         sub_pop=pop.apply_bool_mask(tested, untested_population))
-                # "reset" dummy partner columns
-                pop.set_present_variable(col.NSTP_LAST_TEST, 0,
-                                         sub_pop=pop.apply_bool_mask(tested, untested_population))
-                pop.set_present_variable(col.NP_LAST_TEST, 0,
-                                         sub_pop=pop.apply_bool_mask(tested, untested_population))
+                pop.set_present_variable(col.EVER_TESTED, tested, untested_population)
+                self.apply_test_outcomes_to_sub_pop(pop, tested, untested_population)
 
             if len(prev_tested_population) > 0:
                 # test repeat testers
                 tested = pop.transform_group([col.EVER_TESTED, col.NP_LAST_TEST, col.NSTP_LAST_TEST],
                                              self.calc_testing_outcomes,
                                              sub_pop=prev_tested_population)
-                # set last test date
+                self.apply_test_outcomes_to_sub_pop(pop, tested, prev_tested_population)
+
+    def apply_test_outcomes_to_sub_pop(self, pop, tested, sub_pop):
+        """
+        Uses HIV testing outcomes for a given sub-population to
+        set last test date and reset number of partners since last test.
+        """
+        # set last test date
+        pop.set_present_variable(col.LAST_TEST_DATE, pop.date,
+                                 sub_pop=pop.apply_bool_mask(tested, sub_pop))
+        # "reset" dummy partner columns
+        pop.set_present_variable(col.NSTP_LAST_TEST, 0,
+                                 sub_pop=pop.apply_bool_mask(tested, sub_pop))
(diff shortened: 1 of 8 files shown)


pineapple-cat:
Two ANC tests were implemented (one at the start of the second trimester and the other at the end of the third trimester), in accordance with recently changed SAS ANC testing code. However, `tested_anc` and `tested_labdel` columns haven't been added yet, and there is also no ANC testing for women who are diagnosed but have not disclosed their status.

src/hivpy/hiv_testing.py
```diff
@@ -84,29 +84,70 @@ def update_hiv_testing(self, pop):
...
+        COVID disruption is factored in.
+        """
+        # conduct two tests in anc during pregnancy if there is no covid disruption
+        if not (self.covid_disrup_affected | self.testing_disrup_covid):
+            # get population in second trimester
+            in_second_trimester = pop.get_sub_pop([(col.HIV_STATUS, op.eq, False),
+                                                   (col.ANC, op.eq, True),
+                                                   (col.PREGNANT, op.eq, True),
+                                                   (col.LAST_PREGNANCY_DATE, op.lt, pop.date - timedelta(days=90)),
+                                                   (col.LAST_PREGNANCY_DATE, op.ge, pop.date - timedelta(days=180))])
```
```suggestion
                                                   (col.LAST_PREGNANCY_DATE, op.lt, pop.date - timedelta(days=90)),
                                                   (col.LAST_PREGNANCY_DATE, op.ge, pop.date - time_delta(days=90) + time_step)])
```

Given the SAS code this should only execute once during pregnancy. In the SAS it happens exactly 3 months into pregnancy, but because we are being timestep agnostic we can make it just the first time step which is at least 3 months into pregnancy.

mmcleod89 (commented):
There is an additional complication in the SAS code which doesn't make sense to me and I will email the team about. 
`if ever_tested ne 1 then do; ever_tested=1; date1test=caldate{t}; newp_lasttest_tested_this_per = newp_lasttest; newp_lasttest=0;end;`
This executes after checking for people 3 or 9 months into pregnancy; people 9 at birth definitely get a test but people 3 months in are 50/50. But this happens to everyone who is 3 or 9 months in which I think is inconsistent, so I want to check that their code is doing what they want it to be before we sign off on this version.

mmcleod89:
We might want to also consider resetting ANC=0 when people give birth (as they do in SAS), in which case we don't need to check ANC and pregnancy at the same time since people with ANC=1 will always be pregnant. But it doesn't affect the outcomes right now so not a big deal.

pineapple-cat:
SAS ANC code changes:
- [x] Potential of up to three tests during pregnancy
- [x] Post-delivery test (1 month after birth)
- [x] Make a new issue for: ANC testing for women who are diagnosed but have not disclosed their status

mmcleod89:
1st, 2nd, and 3rd trimester test probabilities need to be added into the code. (May be sampled.)

src/hivpy/hiv_testing.py
```diff
@@ -84,29 +90,87 @@ def update_hiv_testing(self, pop):
...
+                                                   (col.LAST_PREGNANCY_DATE, op.le, pop.date
+                                                    - timedelta(days=270))])
+            self.update_sub_pop_test_date(pop, third_trimester_pop, self.prob_anc_test_trim3)
+
+            # get post-delivery population tested during the previous time step
+            post_delivery_pop = pop.get_sub_pop([(col.HIV_STATUS, op.eq, False),
+                                                 (col.LAST_TEST_DATE, op.eq, pop.date - time_step),
+                                                 (col.LAST_PREGNANCY_DATE, op.le, pop.date
+                                                  - (timedelta(days=270) + time_step))])
+            self.update_sub_pop_test_date(pop, post_delivery_pop, self.prob_test_postdel)
```
mmcleod89: There is a bit of an issue here. This is looking for anyone whose last test date was the previous timestep, and whose last pregnancy was at least (9 months + 1 timestep) ago. But this would apply to _anyone_ who has given birth in the past and is not currently pregnant, even if it were years ago.

mmcleod89 (commented):
The post delivery logic needs a tweak to make sure that it is only executed in the timestep immediately after delivery. I think a check in test to show that no people not in ANC were tested would be nice as well, which can just go in at the end.

mmcleod89 (commented):
Happy with this!

mmcleod89 (approved):

PR #132, opened 2023-06-15
Add VMMC HIV testing to Circumcision Module

### LINKED ISSUE #101: Add HIV testing components to Circumcision Module
To be completed when there is a suitable HIV testing infrastructure
- relies on testing at t-1
- see circumcision intervention in SAS code

pineapple-cat:
Addresses #101.

### DIFF
--- a/src/hivpy/circumcision.py
+++ b/src/hivpy/circumcision.py
@@ -129,7 +129,7 @@ class CircumcisionModule:
                & (self.policy_intervention_year <= self.date.year):
                 uncirc_male_population = pop.get_sub_pop([(col.SEX, op.eq, SexType.Male),
                                                           (col.CIRCUMCISED, op.eq, False),
-                                                          (col.HIV_STATUS, op.eq, False),
+                                                          (col.HIV_DIAGNOSED, op.eq, False),
                                                           (col.HARD_REACH, op.eq, False),
                                                           (col.AGE, op.ge, self.vmmc_cutoff_age),
                                                           (col.AGE, op.lt, self.max_vmmc_age)])
@@ -137,7 +137,7 @@ class CircumcisionModule:
             else:
                 uncirc_male_population = pop.get_sub_pop([(col.SEX, op.eq, SexType.Male),
                                                           (col.CIRCUMCISED, op.eq, False),
-                                                          (col.HIV_STATUS, op.eq, False),
+                                                          (col.HIV_DIAGNOSED, op.eq, False),
                                                           (col.HARD_REACH, op.eq, False),
                                                           (col.AGE, op.ge, self.min_vmmc_age),
                                                           (col.AGE, op.lt, self.max_vmmc_age)])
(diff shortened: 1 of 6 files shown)


src/hivpy/circumcision.py
```diff
@@ -158,6 +162,17 @@ def update_vmmc(self, pop):
...
+                # those that just got circumcised and weren't tested last time step get tested now
+                just_tested = pop.get_sub_pop([(col.CIRCUMCISION_DATE, op.eq, pop.date),
+                                               (col.LAST_TEST_DATE, op.ne, pop.date - time_step)])
+                # correctly set up related columns
+                if len(just_tested) > 0:
+                    pop.set_present_variable(col.EVER_TESTED, True, just_tested)
+                    pop.set_present_variable(col.LAST_TEST_DATE, pop.date, just_tested)
+                    # "reset" dummy partner columns
+                    pop.set_present_variable(col.NSTP_LAST_TEST, 0, just_tested)
+                    pop.set_present_variable(col.NP_LAST_TEST, 0, just_tested)
```
pineapple-cat: This turned out to be a smaller bit of code than I initially expected, but it should probably still be moved to the HIV testing module.

mmcleod89:
We should add `test_link_circ` to `update_vmmc`
- If a person was tested _last timestep_ and not diagnosed with HIV (i.e. test negative) then that person has `test_link_circ_prob` chance of being circumcised
- Only applies to men **under 50**

pineapple-cat:
Making a note here of the `test_link_circ` section in the SAS code:
```
if test_link_circ=1 and t ge 2 and tested_tm1=1 and registd_tm1  ne 1 then do;
    uc=rand('uniform'); if uc < test_link_circ_prob  and vmmc_disrup_covid ne 1 then do; mcirc=1;new_mcirc=1; end;
end;
```

src/hivpy/hiv_testing.py
```diff
@@ -108,6 +108,42 @@ def update_hiv_testing(self, pop):
...
+                pop.set_present_variable(col.CIRCUMCISED, circumcision, tested_uncirc_male_pop)
+                pop.set_present_variable(col.VMMC, circumcision, tested_uncirc_male_pop)
+
+    def update_vmmc_hiv_testing(self, pop, time_step):
+        """
+        Update HIV testing status after VMMC.
+        """
+        # those that just got circumcised and weren't tested last time step get tested now
+        just_tested = pop.get_sub_pop([(col.CIRCUMCISION_DATE, op.eq, pop.date),
+                                       (col.LAST_TEST_DATE, op.ne, pop.date - time_step)])
```
mmcleod89: This seems very specific and might be related to a 3-monthly time step. I suspect that if the timestep were one month then we would actually want to be checking for anyone who has not had a test in the last three months, but we should double check with the team.

pineapple-cat: Looks like this function should work differently. If someone has decided to get VMMC, they are tested first, and only if they are found to be HIV negative (undiagnosed) do they get circumcised.

mmcleod89:
We need to check the filtering of men on HIV status is not done prematurely 
We should update any relevant comments / docs to make it clear that the `prob_circ` is the probably of putting yourself forward for circumcision rather than getting circumcised itself as it may depend on HIV status / diagnosis.

mmcleod89 (approved):

PR #142, opened 2023-07-07
Output / Sex Behaviour refactor 

mmcleod89:
Remove old style data accesses from output & sex behaviour 
Get rid of the old `selector` function which was made redundant by sub-populations

### DIFF
--- a/src/hivpy/simulation.py
+++ b/src/hivpy/simulation.py
@@ -1,10 +1,5 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
-
-from .population import Population
-
 import logging
 import operator
 from datetime import datetime
@@ -16,6 +11,7 @@ import hivpy.column_names as col
 
 from .common import SexType
 from .config import SimulationConfig
+from .population import Population
 
 
 class SimulationOutput:
--- a/src/hivpy/simulation.py
+++ b/src/hivpy/simulation.py
@@ -1,3 +1,10 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+
+from .population import Population
+
 import logging
 import operator
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


src/hivpy/simulation.py
```diff
@@ -39,44 +43,47 @@ def _update_date(self, date):
...
         # Update HIV prevalence by sex
-        men_idx = selector(pop_data, sex=(operator.eq, SexType.Male))
+        men_idx = pop.get_sub_pop([(col.SEX, operator.eq, SexType.Male)])
         self.output_stats["HIV prevalence (male)"][self.step] = (
-            self._ratio(HIV_pos_idx & men_idx, men_idx & over_15_idx))
-        women_idx = selector(pop_data, sex=(operator.eq, SexType.Female))
-        self.output_stats["HIV prevalence (female)"][self.step] = self._ratio(
-            HIV_pos_idx & women_idx, women_idx & over_15_idx)
+            self._ratio(pop.get_sub_pop_intersection(men_idx, over_15_idx),
+                        pop.get_sub_pop_intersection(men_idx, over_15_idx)))
```
pineapple-cat: This ratio is always going to be 1 because we're comparing the length of the same sub-population to itself. Shouldn't we include `HIV_pos_idx` in the first intersection?

src/hivpy/simulation.py
```diff
@@ -39,44 +43,47 @@ def _update_date(self, date):
...
+        women_idx = pop.get_sub_pop([(col.SEX, operator.eq, SexType.Female)])
+        self.output_stats["HIV prevalence (female)"][self.step] = (
+            self._ratio(pop.get_sub_pop_intersection(women_idx, over_15_idx),
+                        pop.get_sub_pop_intersection(women_idx, over_15_idx)))
```
pineapple-cat: Same as above, this ratio is also always going to be 1.

pineapple-cat (approved):
Everything else looks good to me. I just think we need to note down somewhere that we currently don't model HIV in children. If the code does ever expand in that direction, we'd need to make sure we get an intersection of `sex_idx`, `HIV_pos_idx`, and `over_15_idx` to get correct HIV prevalence ratios.

mmcleod89:
That's true; these outputs are really just for demonstration though and so will get overwritten anyway. Ideally I'd like the outputs for different parts of the simulation to be controlled by their respective modules.

PR #143, opened 2023-07-10
Add more descriptive logical expressions for dataframe

mmcleod89:
Instead of using the unannotated lists to create a statement which is implicitly in conjunctive normal form (which is admittedly a touch confusing!), this gives the option of writing logic expressions for e.g. `get_sub_pop` using three kinds of statements:
-  `COND`: for a single condition e.g. `COND(col.AGE, op.gt, 30)`
- `AND`: for a list of conjunctions which are just given as an abritrary list of expressions in the arguments to the constructor e.g. `AND(expr1, expr2, expr3...). No `[]` lists required!
- `OR`: same as `AND` but for disjunctions. 

It's slightly more verbose by virtue of writing `COND`/`AND`/`OR` but more clearer what is going on, and doesn't require you to parse out the `([(...` stuff that we have with the current system of tuples inside lists inside ... 

**Still needs some tests to be written.** 

You can see an example of how to use this framework in the change to sexual_behaviour, but I've temporarily left the old code in as a comment to make the comparison easier to see if you checkout the code.

### DIFF
--- a/src/hivpy/common.py
+++ b/src/hivpy/common.py
@@ -146,10 +146,10 @@ class LogicExpr(ABC):
     """
     Abstract class for representing general logical expressions as applied to a population.
     Derived classes are: COND for individual conditions (e.g. age > 15)
-                         AND for conjunctions 
+                         AND for conjunctions
                          OR for disjunctions
     AND and OR can take a list of LogicExpr which may be of any of the three types to
-    give full freedom to create any logical statement without being beholden to e.g. 
+    give full freedom to create any logical statement without being beholden to e.g.
     conjunctive normal form.
     """
     @abstractmethod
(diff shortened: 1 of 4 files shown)


pineapple-cat (approved):
I like it; I think the additional clarity from specifying `AND`/`OR` will be invaluable for the SAS team, especially when it comes to debugging.

There's still a fair bit of bracket soup at the end of a statement when mixing and matching conjunctions and disjunctions:
```
population.get_sub_pop(AND(COND(col.A, op.ge, val1),
                           COND(col.B, op.lt, val2),
                           OR(COND(col.C, op.eq, val3),
                              COND(col.D, op.eq, val4))))
```
but at least now they're all the same round bracket, so an IDE should easily be able to flag and sort out any bracket-related issues.

I had a quick look around at how some other Python libraries for propositional logic represent their clauses, but couldn't find anything that struck me as a better and easier to understand representation for our purposes.

PR #157, opened 2023-11-13
Add initial viral load and CD4 count

mmcleod89:
- Initialises viral load and CD4 count when HIV is transmitted. 
- Fixes some transmission issues
- Repairs some tests
- Removes some duplicated functions and variables
- Unit tests for initial viral load and HIV stats


src/hivpy/hiv_status.py
```diff
@@ -16,29 +17,51 @@
...
     def __init__(self):
-        self.stp_HIV_rate = {SexType.Male: np.zeros(5),
-                             SexType.Female: np.zeros(5)}  # FIXME
-        self.stp_viral_group_rate = {SexType.Male: np.array([np.zeros(7)]*5),
-                                     SexType.Female: np.array([np.zeros(7)]*5)}
-        # FIXME move these to data file
+        # self.stp_HIV_rate = {SexType.Male: np.zeros(5),
+        #                      SexType.Female: np.zeros(5)}  # FIXME
+        # self.stp_viral_group_rate = {SexType.Male: np.array([np.zeros(7)]*5),
+        #                              SexType.Female: np.array([np.zeros(7)]*5)}
```
pineapple-cat: Shouldn't we just delete these since they've been replaced by `ratio_infected_stp` and `ratio_vl_stp` instead of leaving them commented out here?

pineapple-cat (approved):

src/hivpy/hiv_status.py
mmcleod89: Yes! Good spot I had forgotten about them.

Issue #158, opened 2023-11-14
Outputs for early epidemic tests 


mmcleod89:
Outputs for checking the behaviour of the simulation in the early epidemic (pre-treatment & testing)

See example outputs given by Jenny, but this will mostly focus on things like:
- Number of people in each age bracket
- Number of deaths
- Number of sexual partners w/demographics
- Number of people with HIV w/demographics
- Number of births

Won't include things like testing as we're focusing on the early era. 

- [x] Output module should write these outputs for each timestep into an output file. 
- [x] Also want to be able to make graphs. (May be post processing or automatic.) 
- [x] List desired outputs in config file. 
- [x] Python script for creating comparison plots of single runs with SAS data. (Extended by #171 to allow data aggregation and averages across multiple runs.)

pineapple-cat:
Current list of outputs to include (from 21/11/23 meeting):

- [x] proportion of people aged 15-64 with a long-term partner (`p_ep`)
- [x] proportion of people aged 15-64 with at least one short-term partner (`p_newp_ge1`)
- [x] proportion of people aged 15-64 with at least five short-term partners (`p_newp_ge5`)
- [x] proportion of women giving birth
- [x] of women with HIV giving birth, proportion of children infected
- [x] proportion of circumcised men aged 15-49
- [x] proportion of female sex workers
- [x] proportion of female sex workers with HIV
- [x] proportion of people aged 15-49 with HIV (prevalence)
- [x] measure of new HIV cases in women and men aged 15-24 (incidence)
- [x] measure of new HIV cases in people aged 15-49 and 15-64 (incidence)
- [ ] of the infections in a given time-step, proportion of people that have been infected by a long-term partner (`p_inf_ep`) (relies on #162)
- [x] of the infections in a given time-step, proportion of people that have been infected by someone in the primary infection phase (`p_inf_primary`)
- [x] number of women and men aged 15+ and 20-59 dead (`n_death_2059_w`, `n_death_2059_m`)
- [x] number of people with CD4 counts <200, 200-500, and >500 (`n_cd4_lt200`)
- [x] number of people aged 15+ alive (`n_alive`)
- [x] number of people with HIV (`n_hiv`)
- [x] number of live births
- [x] number of births to infected women

Additional outputs for validation:

- [x] short-term partner gender balance (`log_gender_r_newp`)
[...]

PR #161, opened 2023-11-17
Simulation output enhancements for early epidemic statistics

pineapple-cat:
Addresses #158 (but not fully, as long-term partner infections are yet to be added).


pineapple-cat:
- [x] `FIXME`: After going through some model outputs, it looks like `date_HIV_infection` changes every time step to match the current date for each infected person. It should remain at the date of the time step during which each individual initially contracted HIV, otherwise we can't rely on this date to determine whether someone is in their primary infection phase or not.

pineapple-cat:
- [x] `FIXME`: Make output its own module to avoid having to pass it down (simulation -> population -> sub-modules).

mmcleod89 (approved):

PR #165, opened 2023-12-06
Fix hiv status tests

mmcleod89:
HIV status tests failing due to new primary infection variable not being set in them, so needs to go into your branch rather than development. 

A couple of other tests are failing; I wouldn't worry about death because that's all getting rewritten anyway, but I don't know why the pregnancy test is failing.


pineapple-cat (approved):
Thank you for finding and fixing the issue. I think pregnancy could just be failing due to unlucky runs/seed value (it doesn't fail when I run it locally), so it should hopefully be fine?

Issue #147, opened 2023-08-22
time step


andrew-phillips-1:
Can we have date as a float and update by 1/12 in each time step, rather than use dates and days ?  That worked OK when I did it in the SAS program.

PR #166, opened 2023-12-11
Add dying

mmcleod89:
A few things happening here:
- HIV related death and illness added for ART naive people
- Non HIV death for HIV+ and HIV- people is updated
- People are dropped from the dataframe after death (but not recording death yet; we should discuss how we want to record this)
- In order to make the maths more sensible for the rates of death I have implemented the Andrew's suggestion to use dates which are just integer months rather than literal dates. This doesn't impact other areas of code as for backwards compatibility I have written my own `date` and `timedelta` classes for now. We can potentially auto-replace these with other names to avoid confusion and clashes later, the output code is merged onto this branch as well. 

**This still needs tests before moving onto the main but could provide a basis for a demo branch including death if combined with the output development branch**


src/hivpy/column_names.py
```diff
@@ -36,43 +36,53 @@
...
-MAX_CD4 = "max_cd4"                             # float: Maximum CD4 count to which a person can return when on ART
+CD4 = "cd4"                                     # None | float: CD4 count per cubic millimeter; set to None for people w/o HIV
+MAX_CD4 = "max_cd4"                             # float: maximum CD4 count to which a person can return when on ART
 ART_NAIVE = "art_naive"                         # bool: True if person has never been on antiretroviral therapy
-X4_VIRUS = "x4_virus"                           # bool: True if X4 virus is present in person, false otherwise.
-
-ADC = "AIDS_defining_condition"                 # Bool: presence of AIDS defining condition
+X4_VIRUS = "x4_virus"                           # bool: True if X4 virus is present in person, False otherwise
 
 DATE_OF_DEATH = "date_of_death"                 # None | datetime.date: date of death if dead, o/w None
```
pineapple-cat, line 70: Do we still need this for something or is it a superfluous column now that dead people are being dropped from the population dataframe?

src/hivpy/common.py
```diff
@@ -8,7 +8,7 @@
 if TYPE_CHECKING:
     from .population import Population
 
-import datetime
+# import datetime
```
pineapple-cat, line 11: Leaving a note here as a reminder to remove leftover commented code (esp imports) before merging to development.
```suggestion
```

src/hivpy/common.py
```diff
@@ -61,8 +61,114 @@ def opposite_sex(sex: SexType):
     return (1 - sex)
 
 
-def diff_years(date_begin, date_end):
-    return (date_end - date_begin) / datetime.timedelta(days=365.25)
+class date:
```
pineapple-cat, line 64: Closes #147.

src/hivpy/population.py
```diff
@@ -286,6 +286,17 @@ def evolve(self, time_step: datetime.timedelta):
...
             self.hiv_testing.update_hiv_testing(self)
+            HIV_deaths = self.hiv_status.HIV_related_disease_risk(self, time_step)
+            n_deaths = n_deaths + sum(HIV_deaths)
+            if (n_deaths and self.apply_death):
+                self.drop_from_population(HIV_deaths)
+
+        # Apply non-hiv deaths
+        non_HIV_deaths = self.demographics.determine_deaths(self, time_step)
+        n_deaths = n_deaths + sum(non_HIV_deaths)
+        # self.data.loc[died_this_period, col.DATE_OF_DEATH] = self.date
```
pineapple-cat, line 297:
```suggestion
```

pineapple-cat (approved):
Curious to see if we'll get much of a performance increase from dropping the dead from the population. I didn't double-check the exact values used, but there's nothing that glaringly stands out to me as wrong from the death and disease additions compared to their SAS code counterparts (or what I thought I could find of them).

Since the `Demo_12_23` branch contains some glue code for the dying and output additions, we should probably merge our branches there first and then pass on the smoothed over changes to `development`.

src/hivpy/column_names.py
mmcleod89: Now dropped in Demo branch

PR #170, opened 2024-01-15
PyCharm setup and package installation tutorial

### LINKED ISSUE #149: Python & PyCharm Settings Tutorial
Add a tutorial for setup of various settings, including something like creating and using a venv. Perhaps with side-by-side steps using both terminal commands and the PyCharm interface?

pineapple-cat:
Addresses #149.

### DIFF
--- a/tutorials/setup.md
+++ b/tutorials/setup.md
@@ -1,0 +1,29 @@
+## Setup and Package Installation with PyCharm
+
+The following steps outline how to set up a new, clean installation of the HIVpy package. Instruction specifics are based on the assumption that PyCharm v2023 is used.
+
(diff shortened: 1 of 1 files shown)


mmcleod89:
- Add instructions for opening a terminal before checking venv / running terminal commands
- Add that you may have to go to "VSC" -> "get from version control" if there is no "git" in the menu

pineapple-cat:
- Add a step for opening the project if it doesn't open automatically.
- Add that venv creation defaults should be fine, but users need to ensure Python v3.7+ is being used.
- Add reminder of how users can tell their venv is running.

mmcleod89 (approved):

PR #173, opened 2024-01-29
Early epidemic outputs compared to December Demo

mmcleod89:


src/hivpy/output.py
```diff
@@ -231,12 +233,14 @@ def _update_partner_sex_balance(self, pop: Population):
         men_stp_age_list = (np.concatenate(men_stp_age_list).ravel() if len(men_stp_age_list) > 0
                             else men_stp_age_list).tolist()
 
+        # FIXME: should we log all ratios here or have this step happen in post?
+        # NOTE: sum type converted from numpy.int64
         self.output_stats.loc[self.step, "Partner sex balance (male)"] = self._log(
-            self._ratio(pop.get_variable(col.NUM_PARTNERS, active_men).sum(),
-                        pop.get_variable(col.NUM_PARTNERS, active_women).sum()))
+            self._ratio(int(pop.get_variable(col.NUM_PARTNERS, active_men).sum()),
+                        int(pop.get_variable(col.NUM_PARTNERS, active_women).sum())))
```
mmcleod89, line 240: Why is this conversion necessary?

pineapple-cat: I couldn't get `test_partner_sex_balance` to work without it due to the changes I made to `_ratio` - it expects either integers or something you can get a length for (like a pandas df or series), and the numpy int type doesn't pass the `if type(pop) is int` condition.

PR #182, opened 2024-04-15
HIV testing for the symptomatic

### LINKED ISSUE #111: HIV testing for Symptomatic people (without diagnosis)
SAS lines 7369-7389

People with AIDS defining condition / TB etc. may get tested separately from other testing considerations.

pineapple-cat:
Addresses #111.


src/hivpy/hiv_testing.py
```diff
@@ -50,81 +55,125 @@ def __init__(self, **kwargs):
...
+                          (s < self.prob_test_non_hiv_symptoms))
+                # set outcomes
+                pop.set_present_variable(col.TEST_MARK, marked, not_diag_tested_pop)
+
+    def test_mark_vmmc(self, pop):
+        """
+        Mark recently circumcised individuals to undergo testing this time step.
+        """
+        # those that just got circumcised and weren't tested last time step get tested now
+        # FIXME: should the actual time step be used here instead of a flat 90 days?
```
pineapple-cat: Forgot to raise this question earlier but making a note here to ask @UCL/hiv-modelling during our meeting tomorrow.

pineapple-cat: Meeting notes:
- Use `time_step` instead of `timedelta(days=90)` when selecting for VMMC testing.
- Move VMMC update after HIV testing update so people about to undergo VMMC are tested first.
  - Adheres more logically to reality.
  - Allows us to rely on testing and diagnosis for the current time step.
  - Indirectly introduces a healing period before VMMC affects sexual behaviour.

pineapple-cat: Opened issue #185 for the second point and implemented the first.

src/hivpy/hiv_testing.py
```diff
@@ -50,81 +55,125 @@ def __init__(self, **kwargs):
...
+        Mark HIV symptomatic individuals to undergo testing this time step.
+        """
+        # testing occurs after a certain year
+        if (pop.date.year >= self.date_start_testing):
+
+            # update symptomatic test probabilities
+            if pop.date.year <= 2015:
+                self.prob_test_who4 = min(0.9, self.prob_test_who4 * self.incr_test_rate_sympt)
+                self.prob_test_tb = min(0.8, self.prob_test_tb * self.incr_test_rate_sympt)
+                self.prob_test_non_tb_who3 = min(0.7, self.prob_test_non_tb_who3 * self.incr_test_rate_sympt)
```
mmcleod89: This has a potential issue because it relies on the cumulative effects of `inc_test_rate_sympt` each timestep, and presume that a simulation _must_ start before `date_start_testing`. I expect that we would always run from before this date, but it is important to note that this code will not work if the run starts after that date, and at the very least that needs to be clearly documented somewhere.

src/hivpy/hiv_testing.py
```diff
@@ -50,81 +55,125 @@ def __init__(self, **kwargs):
...
+
+            # undiagnosed and untested population
+            not_diag_tested_pop = pop.get_sub_pop([(col.HIV_DIAGNOSED, op.eq, False),
+                                                   (col.EVER_TESTED, op.eq, False),
+                                                   (col.TEST_MARK, op.eq, False)])
```
```suggestion
            not_diag_tested_pop = pop.get_sub_pop([(col.HIV_DIAGNOSED, op.eq, False),
                                                   (col.TEST_MARK, op.eq, False)])
```

I don't think that this should require the person to never have been tested, just that they haven't been tested for any other reason this timestep. See SAS code:
`if t ge 3 and registd ne 1 and tested ne 1 and caldate{t} ge date_start_testing  then do; `

src/hivpy/hiv_testing.py
```diff
@@ -50,81 +55,125 @@ def __init__(self, **kwargs):
...
-        # testing occurs after a certain year if there is no covid disruption
-        if ((pop.date.year >= self.date_start_testing)
-           & (not (self.covid_disrup_affected | self.testing_disrup_covid))):
+        # mark people for testing
+        # hiv symptomatic > non hiv symptomatic > vmmc > anc > self testing > general > prep
+        self.test_mark_hiv_symptomatic(pop)
+        self.test_mark_non_hiv_symptomatic(pop)
+        self.test_mark_vmmc(pop, time_step)
+        self.test_mark_anc(pop, time_step)
+        self.test_mark_general_pop(pop)
```
mmcleod89: In the SAS code the order appears to be:
1. Test for non HIV TB (doesn't appear to exist in HIVpy yet)
2. Circumcision related testing
3. Non HIV symptoms
4. ANC
5. Post Delivery
6. General population (including sex workers)
7. Testing before prep (not in HIVpy yet)
8. routine testing on prep (not in HIVpy yet)
9. HIV symptomatic testing

src/hivpy/hiv_testing.py
```diff
@@ -50,81 +55,125 @@ def __init__(self, **kwargs):
...
-            tested = r < prob_test
-            pop.set_present_variable(col.LAST_TEST_DATE, pop.date,
-                                     sub_pop=pop.apply_bool_mask(tested, sub_pop))
+        # testing occurs after a certain year if there is no covid disruption
+        if ((pop.date.year >= self.date_start_testing)
+           & (not (self.covid_disrup_affected | self.testing_disrup_covid))):
+
+            # undiagnosed (last time step) and untested population
+            not_diag_tested_pop = pop.get_sub_pop([(pop.get_correct_column(col.HIV_DIAGNOSED, dt=1), op.eq, False),
+                                                   (col.EVER_TESTED, op.eq, False),
```
mmcleod89: Similar to above, we don't need EVER_TESTED false here.

src/hivpy/hiv_testing.py
```diff
@@ -50,81 +55,125 @@ def __init__(self, **kwargs):
...
+                                                   (col.TEST_MARK, op.eq, False)])
 
-    def update_anc_hiv_testing(self, pop, time_step):
+            if len(not_diag_tested_pop) > 0:
+                # mark people for testing
+                r = rng.uniform(size=len(not_diag_tested_pop))
+                s = rng.uniform(size=len(not_diag_tested_pop))
+                marked = ((r < (self.prob_test_non_tb_who3 + self.prob_test_who4)/2) &
+                          (s < self.prob_test_non_hiv_symptoms))
```
mmcleod89, line 161: I understand that `r` is related to the SAS line `u < (test_rate_non_tb_who3+test_rate_who4)/2 then do; `, but what does `s` /  `prob_test_non_hiv_symptoms` correspond to?

src/hivpy/hiv_testing.py
```diff
@@ -145,67 +194,77 @@ def update_anc_hiv_testing(self, pop, time_step):
...
+            self.rate_first_test = self.init_rate_first_test + (min(pop.date.year, self.date_test_rate_plateau)
+                                                                - self.date_start_testing) \
+                                                                * self.an_lin_incr_test
+            self.rate_rep_test = (min(pop.date.year, self.date_test_rate_plateau)
+                                  - self.date_start_testing) * self.an_lin_incr_test
+
+            # general population ready for testing
+            testing_population = pop.get_sub_pop([(col.HARD_REACH, op.eq, False),
+                                                  (col.AGE, op.ge, 15),
+                                                  (col.HIV_STATUS, op.eq, False),
```
```suggestion
                                                  (col.HIV_DIAGNOSED, op.eq, False),
```
needs to be diagnosed, not HIV status, because otherwise no HIV+ people will get tested!

src/hivpy/hiv_testing.py
```diff
@@ -145,67 +194,77 @@ def update_anc_hiv_testing(self, pop, time_step):
...
+            marked = r < prob_test
+            # set outcomes
+            pop.set_present_variable(col.TEST_MARK, True, sub_pop=pop.apply_bool_mask(marked, sub_pop))
+
+    def test_mark_general_pop(self, pop):
+        """
+        Mark general population to undergo testing this time step.
+        """
+        # testing occurs after a certain year if there is no covid disruption
+        if ((pop.date.year >= self.date_start_testing)
```
mmcleod89: SAS code features `caldate{t} ge (date_start_testing+5.5)` for this section, should this date be pushed later?

src/hivpy/hiv_testing.py
```diff
@@ -145,67 +194,77 @@ def update_anc_hiv_testing(self, pop, time_step):
...
+            self.rate_rep_test = (min(pop.date.year, self.date_test_rate_plateau)
+                                  - self.date_start_testing) * self.an_lin_incr_test
+
+            # general population ready for testing
+            testing_population = pop.get_sub_pop([(col.HARD_REACH, op.eq, False),
+                                                  (col.AGE, op.ge, 15),
+                                                  (col.HIV_STATUS, op.eq, False),
+                                                  [(col.LAST_TEST_DATE, op.le, pop.date -
+                                                    timedelta(days=self.days_to_wait[self.eff_max_freq_testing])),
+                                                   (col.LAST_TEST_DATE, op.eq, None)],
```
mmcleod89: Maybe give this an explicit `OR` syntax to make it easier to read?

src/hivpy/hiv_testing.py
```diff
@@ -145,67 +194,77 @@ def update_anc_hiv_testing(self, pop, time_step):
...
+                                             self.calc_testing_outcomes, sub_pop=testing_population)
+                # set outcomes
+                pop.set_present_variable(col.TEST_MARK, marked, testing_population)
+
+    def calc_testing_outcomes(self, repeat_tester, np_last_test, nstp_last_test, size):
+        """
+        Uses the HIV test probability for either
+        first-time or repeat testers to return testing outcomes.
+        """
+        prob_test = self.calc_prob_test(repeat_tester, np_last_test, nstp_last_test)
```
mmcleod89, line 262: This function `calc_prob_test` doesn't do anything. I think this needs to be fixed?

mmcleod89 (commented):
Sex workers should also be added to the general population testing 

```
if t ge 2 and sw_test_6mthly=1 and sw=1 and (caldate{t}-dt_last_test >= 0.5 or dt_last_test=.) then do;
		tested=1; tested_as_sw=1; 
		if ever_tested ne 1 then date1test=caldate{t}; ever_tested=1; dt_last_test=caldate{t}; 
		np_lasttest=0; newp_lasttest_tested_this_per=newp_lasttest; newp_lasttest=0;
	end;
```

src/hivpy/hiv_testing.py
```diff
@@ -202,6 +202,8 @@ def test_mark_anc(self, pop, time_step):
                                                    (col.LAST_PREGNANCY_DATE, op.le, pop.date
                                                     - timedelta(days=270))])
             self.update_sub_pop_test_date(pop, third_trimester_pop, self.prob_anc_test_trim3)
+            # remove from antenatal care
+            pop.set_present_variable(col.ANC, False, third_trimester_pop)
```
mmcleod89: Is there somewhere we can put this that is more intuitive? It's not really a part of the testing logic per se

src/hivpy/hiv_testing.py
mmcleod89: I'm not sure how the discussion of VMMC vs HIV ordering affects this precisely, but it does seem like the HIV symptomatic needs tobe moved down to the bottom

pineapple-cat: I can't remember when we had this discussion, but at some point I asked the HIV team for a testing order and they gave me the following: HIV symptomatic > non-HIV symptomatic > VMMC > ANC > self testing > general > PrEP. That's what the current test mark ordering is based on. If you think something doesn't make sense, we can revisit this during our next meeting.

src/hivpy/hiv_testing.py
pineapple-cat: Three lines above: `if s < rate_non_hiv_symptoms then do;`.

src/hivpy/hiv_testing.py
pineapple-cat: I think we may have included that +5.5 in `date_start_testing` itself by now, but we should double-check.

src/hivpy/hiv_testing.py
pineapple-cat: It's not, but I couldn't think of a clearer placement without doing more writing. This needs to happen after women in ANC give birth, but the ANC testing code needs to make use of their ANC status first, only then can they be discharged from ANC. I suppose I could make a cleanup function in pregnancy for this purpose and then have it called in the population's `evolve()` after testing?

src/hivpy/hiv_testing.py
pineapple-cat: The more I use this syntax, the less I like having to declare each `COND()` separately. How would you feel about just passing condition tuples into each `AND()` and `OR()` and then having them internally convert the tuples into `COND`s? I think it would make sub-pop selections look nicer and get them to take up less space.

src/hivpy/hiv_testing.py
mmcleod89: @andrew-phillips-1 What is the difference between the use of `date_start_testing` and `date_start_testing+5.5` in the SAS code? Crucially, I don't think we can include the +5.5 in `date_start_testing` as noted above because it would change the pattern of increasing testing which needs to start at the earlier date in the SAS code:
```
if date_start_testing lt caldate{t} le 2015  then do; 
	test_rate_who4 = min(0.9,test_rate_who4*incr_test_rate_sympt);  
	test_rate_tb  = min(0.8,test_rate_tb*incr_test_rate_sympt);  
	test_rate_non_tb_who3 = min(0.7,test_rate_non_tb_who3*incr_test_rate_sympt); * 0.7 mar19;
* testing for hiv for a person with non_hiv_tb (i.e. who was hiv negative in last period) ;  * update_24_4_21;
	*This is intended for people who did not have HIV 3 months ago;																																																
	if caldate{t} - date_last_non_hiv_tb = 0.25 and tested ne 1 then do;   * ts1m - dependent on time step ;
		e=rand('uniform'); 
		if e < test_rate_tb then do;
			tested=1; 
			if ever_tested ne 1 then date1test=caldate{t}; ever_tested=1; dt_last_test=caldate{t}; 
			np_lasttest=0; newp_lasttest_tested_this_per=newp_lasttest; newp_lasttest=0;
		end;
	end;
end;
```

src/hivpy/hiv_testing.py
mmcleod89: I would prefer a cleanup function to be called from evolve!

src/hivpy/hiv_testing.py
```diff
@@ -49,163 +55,243 @@ def __init__(self, **kwargs):
...
-            r = rng.uniform(size=len(sub_pop))
-            tested = r < prob_test
-            pop.set_present_variable(col.LAST_TEST_DATE, pop.date,
-                                     sub_pop=pop.apply_bool_mask(tested, sub_pop))
+        # testing occurs after a certain year if there is no covid disruption
+        if ((pop.date.year >= self.date_start_testing)
+           & (not (self.covid_disrup_affected | self.testing_disrup_covid))):
+
+            # undiagnosed (last time step) population not scheduled for testing
+            not_diag_tested_pop = pop.get_sub_pop([(pop.get_correct_column(col.HIV_DIAGNOSED, dt=1), op.eq, False),
```
pineapple-cat: No particular reason to use diagnosis of HIV from the last time step here, diagnosis has not yet been updated and it would be helpful to not need to track multiple time steps worth of it. If HIV diagnosis date is relevant, there is a separate column for this information.

src/hivpy/hiv_testing.py
andrew-phillips-1: Thanks for pointing this out Michael @mmcleod89 .  The point about the 5.5 is that we wanted the rate of testing to start increasing from 5.5 years after testing introduction in 2003.5.  So having been constant until 2009 the rate starts to increase until date_test_rate_plateau.

But I can see that in this line:
rate_1sttest = initial_rate_1sttest + (min(caldate{t},date_test_rate_plateau)-(date_start_testing+5.5))*an_lin_incr_test;

we need to ensure that the second term is not negative, which I think it currently can be.

We don't need the loop starting: if date_start_testing lt caldate{t} le 2015  then do; 

because we don't change the date of start of testing and it is well before 2015.

But I'm not sure if this is answering your question ?

src/hivpy/hiv_testing.py
mmcleod89: Push ANC / post delivery above non-hiv symptoms

src/hivpy/hiv_testing.py
mmcleod89: We need:

- date_testing_introduced
- date for general population testing growth (date_general_testing_growth ?)
- date for general population testing plateau (date_general_testing_plateau ?)
- date for specific testing (symptomatic, ANC etc.) testing plateau

mmcleod89: General population plateau date is sampled, the specific testing plateau is 2015. 
Date testing introduced is 2003.5 and the date for general population testing growth is 2009

src/hivpy/pregnancy.py
```diff
@@ -154,6 +154,7 @@ def update_antenatal_care(self, pop: Population):
         # get population that became pregnant this time step
         pregnant_population = pop.get_sub_pop([(col.PREGNANT, op.eq, True),
                                                (col.LAST_PREGNANCY_DATE, op.eq, pop.date)])
+        # FIXME: should be affected by date_start_testing and date_specific_testing_plateau
```
mmcleod89: How does this relate to the testing dates? Is there an issue for this?

pineapple-cat: I don't think we have an issue for this yet, but I noticed that in the SAS code ANC probability determination looks like this:
```
* RATE OF ATTENDING AN ANC FOR PREGNANT WOMEN; 

if gender=2 then do;
	if      date_start_testing le caldate{t} lt 2015    then prob_anc      = max(prob_anc, 0.1)+rate_anc_inc; * dependent_on_time_step_length ;
	if                            caldate{t} =  2014.75 then prob_anc_2015 = prob_anc;
	if                      	  caldate{t} ge 2015    then prob_anc      = prob_anc_2015;
	if prob_anc gt 0.975   then prob_anc=0.975;  
```
I can't remember if it was always like this, but you can see that there's a different ANC attendance probability before and after 2015.

mmcleod89:
Pushed 2 changes:

1. I think "targeted" is better than "specific", because to me it conveys what's happening a bit better, but you can unroll this change if you disagree. 
2. I think there was an issue with the pregnancy test: it used to assign a random number of people (roughly half) to not want children. Then it used the number of people who _don't_ want children to calculate the expectation for the number of people who would be pregnant, which would work if it were exactly half but in general it won't be. I've removed the uncertaintly by making it exactly half, and for the sake of correctness using `N - num_not_want_children`

mmcleod89:
Are you happy for me to do a final check over and then merge in this afternoon?

pineapple-cat:
> Are you happy for me to do a final check over and then merge in this afternoon?

No objections; I think this PR is in a reasonable state and we'll fix new issues as we discover them.

src/hivpy/hiv_testing.py
```diff
@@ -49,163 +55,240 @@ def __init__(self, **kwargs):
...
+                                                   (col.TEST_MARK, op.eq, False)])
+
+            if len(not_diag_tested_pop) > 0:
+                # mark people for testing
+                marked = pop.transform_group([pop.get_correct_column(col.ADC, dt=1),
+                                              pop.get_correct_column(col.TB, dt=1),
+                                              pop.get_correct_column(col.TB, dt=2),
+                                              pop.get_correct_column(col.NON_TB_WHO3, dt=1)],
+                                             self.calc_symptomatic_testing_outcomes,
+                                             sub_pop=not_diag_tested_pop)
```
mmcleod89: I know I'm risking dragging this PR out further, but I do want to check this behaviour: we're looking at the presence of TB in the last two timesteps, but how does this translate to a different timestep e.g. 1 month? Does this need to be changed? @andrew-phillips-1

pineapple-cat: Relevant test probability determination code for context:
```
if t ge 3 and registd ne 1 and tested ne 1 and caldate{t} ge date_start_testing  then do; 
		if adc_tm1=1                                     then do; rate_test=test_rate_who4; elig_test_who4=1; end; *rate_test*incr_diag_rate_who4;			
		if adc_tm1=0 and tb_tm2 =0 and tb_tm1 =1         then do; rate_test=test_rate_tb; elig_test_tb=1; end;*rate_test*incr_diag_rate_tb;
		if non_tb_who3_ev_tm1 =1 and adc_tm1=0 and tb_tm1=0 then do; rate_test=test_rate_non_tb_who3; elig_test_non_tb_who3=1; end; *rate_test*incr_diag_rate_non_tb_who3;

		if (adc_tm1=1 or (tb_tm2 =0 and tb_tm1 =1) or non_tb_who3_ev_tm1 =1) then do;
			unitest=rand('uniform');if unitest<rate_test and ( (testing_disrup_covid ne 1 or covid_disrup_affected ne 1)) then do;
				tested=1;
				if ever_tested ne 1 then date1test=caldate{t}; ever_tested=1; dt_last_test=caldate{t}; 
				np_lasttest=0; newp_lasttest_tested_this_per=newp_lasttest; newp_lasttest=0;
				sympt_diag=1; sympt_diag_ever=1;
				if adc_tm1=1 then elig_test_who4_tested=1; if adc_tm1=0 and tb_tm2=0 and tb_tm1 =1 then elig_test_tb_tested=1;
				if non_tb_who3_ev_tm1 =1 and adc_tm1=0 and tb_tm1 =0 then elig_test_non_tb_who3_tested=1;
```

andrew-phillips-1: Yes I think the intent is that a person would test for HIV when first diagnosed for TB.  I think since TB treatment lasts 6 months we originally had have tb=1 for two consecutive periods, but I am not seeing that in the code now.  I think that was the reason for specifying the tb_tm2 = 0 - to avoid a person being tested twice for HIV during the TB episode.  Perhaps we no longer need that if TB is only lasting one period.

mmcleod89: So do we only want to have a chance of testing based on TB in the first time step after acquiring TB, and then not for subsequent timesteps? Or, if we have a one month timestep, should we have some possibility of being testing throughout that six month period?

andrew-phillips-1: I think it's enough to test once only for each TB episode in a person without diagnosed HIV

src/hivpy/pregnancy.py
pineapple-cat: Should be addressed by issue #189.

src/hivpy/hiv_testing.py
```diff
@@ -99,46 +102,41 @@ def test_mark_hiv_symptomatic(self, pop):
...
                 # mark people for testing
-                marked = pop.transform_group([pop.get_correct_column(col.ADC, dt=1),
-                                              pop.get_correct_column(col.TB, dt=1),
-                                              pop.get_correct_column(col.TB, dt=2),
-                                              pop.get_correct_column(col.NON_TB_WHO3, dt=1)],
+                marked = pop.transform_group([col.ADC, col.TB_INFECTION_DATE, col.NON_TB_WHO3],
                                              self.calc_symptomatic_testing_outcomes,
                                              sub_pop=not_diag_tested_pop)
                 # set outcomes
                 pop.set_present_variable(col.TEST_MARK, marked, not_diag_tested_pop)
```
pineapple-cat: Not sure we can actually use `col.TB_INFECTION_DATE` in `transform_group`; the date is unhashable and it looks like the DataFrame's `groupby` won't work with it.

src/hivpy/hiv_status.py
```diff
@@ -390,6 +391,8 @@ def disease_and_diagnosis(disease_col, diagnosis_col, disease_rate, diagnosis_pr
         pop.set_present_variable(col.WHO3_EVENT, (who3_disease | tb), HIV_pos)
         new_tb = pop.get_sub_pop_from_array(tb, HIV_pos)
         pop.set_present_variable(col.TB_INFECTION_DATE, pop.date, new_tb)
+        pop.set_present_variable(col.TB_PRIMARY_INFECTION, False)  # FIXME: move this reset elsewhere?
+        pop.set_present_variable(col.TB_PRIMARY_INFECTION, True, new_tb)
```
pineapple-cat: There should be somewhere better to put the reset of TB primary infection status; thoughts?

src/hivpy/hiv_testing.py
```diff
@@ -49,163 +55,238 @@ def __init__(self, **kwargs):
...
+            marked = r < prob_test
+            # set outcomes
+            pop.set_present_variable(col.TEST_MARK, True, sub_pop=pop.apply_bool_mask(marked, sub_pop))
+
+    def test_mark_general_pop(self, pop):
+        """
+        Mark general population to undergo testing this time step.
+        """
+        # testing occurs after a certain year if there is no covid disruption
+        if ((pop.date.year >= self.date_start_testing)
```
```suggestion
        if ((pop.date.year >= self.date_rate_testing_incr)
```

src/tests/test_sexual_behaviour.py
```diff
@@ -235,7 +235,7 @@ def test_risk_adc():
     init_HIV_idx = rng.integers(0, N, size=5)
     init_ADC_idx = init_HIV_idx[[0, 2, 4]]
     pop.data.loc[init_HIV_idx, col.HIV_STATUS] = True
-    pop.data.loc[init_ADC_idx, col.ADC] = True
+    pop.data.loc[init_ADC_idx, pop.get_correct_column(col.ADC, dt=0)] = True
```
mmcleod89:
```suggestion
    pop.data.loc[init_ADC_idx, col.ADC] = True
```

src/tests/test_sexual_behaviour.py
```diff
@@ -247,7 +247,7 @@ def test_risk_adc():
     add_HIV_idx = rng.integers(0, N, size=5)
     add_ADC_idx = add_HIV_idx[[0, 1, 2]]
     pop.data.loc[add_HIV_idx, col.HIV_STATUS] = True
-    pop.data.loc[add_ADC_idx, col.ADC] = True
+    pop.data.loc[add_ADC_idx, pop.get_correct_column(col.ADC, dt=0)] = True
```
mmcleod89:
```suggestion
    pop.data.loc[add_ADC_idx, col.ADC] = True
```

src/tests/test_hiv_testing.py
```diff
@@ -9,15 +9,29 @@
...
 
+
+def test_hiv_testing_before_start():
+
+    # build population
+    N = 100
+    pop = Population(size=N, start_date=date(2003, 1, 1))
+    pop.data[col.ADC] = True
+    # start date is before testing begins
+    pop.hiv_testing.date_start_testing = 2003.5
```
pineapple-cat: One thing to note about the testing start date is that it's a float and we currently only compare the population date year against it (months and days not taken into account) due to comparison incompatibilities between dates and floats. This is an issue across multiple different modules, not just this one, and should probably be addressed separately from this PR.

mmcleod89: Why not just use a date instead of a float then?

mmcleod89 (approved):
Hooray!

PR #187, opened 2024-06-03
HIV diagnosis for people in primary infection and general population

### LINKED ISSUE #112: HIV diagnosis from testing
- Multiple types of test, 3rd gen most important as this is the default 
- "primary" refers to testing recent HIV infection (within 3 months) with lower sensitivity
- "registd" in SAS code means diagnosed HIV positive (change to e.g. "diagnosed")
- Primary infection testing lines 6897-7053
- Ordinary testing lines 7391-7440
- Tests are not 100% accurate 
- Depends on CD4 count (related to HIV progression) 
- RNA testing for people on CAB PrEP?

### LINKED ISSUE #181: Probability of loss at diagnosis to be implemented with diagnosis code
`eff_prob_loss_at_diag`

In the section about testing those with symptoms (but no diagnosis) there is an additional effective value

`e_eff_prob_loss_at_diag` 

which is used for sex worker specific code.

pineapple-cat:
Addresses #112 and #181.

### DIFF
--- a/src/hivpy/hiv_diagnosis.py
+++ b/src/hivpy/hiv_diagnosis.py
@@ -5,12 +5,14 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from .population import Population
 
+import importlib.resources
 import operator as op
 from enum import IntEnum
 
 import hivpy.column_names as col
 
 from .common import rng
+from .hiv_diagnosis_data import HIVDiagnosisData
 from .prep import PrEPType
 
 
@@ -24,23 +26,28 @@ class HIVTestType(IntEnum):
 class HIVDiagnosisModule:
 
     def __init__(self, **kwargs):
-        # FIXME: move these to data file
+
+        # init hiv diagnosis data
+        with importlib.resources.path("hivpy.data", "hiv_diagnosis.yaml") as data_path:
+            self.hd_data = HIVDiagnosisData(data_path)
+
+        # FIXME: consider how to move more variables to data file
         self.hiv_test_type = HIVTestType.Ab
-        self.init_prep_inj_na = rng.choice([True, False], p=[0.5, 0.5])
-        self.prep_inj_na = rng.choice([True, False], p=[0.5, 0.5]) if self.init_prep_inj_na else 0
+        self.init_prep_inj_na = rng.choice([True, False])
+        self.prep_inj_na = rng.choice([True, False]) if self.init_prep_inj_na else 0
 
-        self.test_sens_general = 0.98
-        self.test_sens_primary_ab = rng.choice([0.5, 0.75])
-        self.test_sens_prep_inj_primary_ab = rng.choice([0, 0.1])
-        self.test_sens_prep_inj_3m_ab = rng.choice([0, 0.2])
-        self.test_sens_prep_inj_ge6m_ab = rng.choice([0.10, 0.25, 0.50])
-        self.tests_sens_prep_inj = rng.choice([0, 1, 2, 3])
-        self.test_sens_prep_inj_primary_na = [0.7, 0.5, 0.3, 0.2][self.tests_sens_prep_inj]
-        self.test_sens_prep_inj_3m_na = [0.85, 0.7, 0.5, 0.3][self.tests_sens_prep_inj]
-        self.test_sens_prep_inj_ge6m_na = [0.95, 0.8, 0.7, 0.5][self.tests_sens_prep_inj]
+        self.test_sens_general = self.hd_data.test_sens_general
+        self.test_sens_primary_ab = self.hd_data.test_sens_primary_ab.sample()
+        self.test_sens_prep_inj_primary_ab = self.hd_data.test_sens_prep_inj_primary_ab.sample()
+        self.test_sens_prep_inj_3m_ab = self.hd_data.test_sens_prep_inj_3m_ab.sample()
+        self.test_sens_prep_inj_ge6m_ab = self.hd_data.test_sens_prep_inj_ge6m_ab.sample()
+        self.tests_sens_prep_inj = self.hd_data.tests_sens_prep_inj.sample()
+        # test_sens_prep_inj used as index for further sensitivity selection
+        self.test_sens_prep_inj_primary_na = self.hd_data.test_sens_prep_inj_primary_na[self.tests_sens_prep_inj]
+        self.test_sens_prep_inj_3m_na = self.hd_data.test_sens_prep_inj_3m_na[self.tests_sens_prep_inj]
+        self.test_sens_prep_inj_ge6m_na = self.hd_data.test_sens_prep_inj_ge6m_na[self.tests_sens_prep_inj]
 
-        self.prob_loss_at_diag = rng.choice([0.02, 0.05, 0.15, 0.35, 0.50],
-                                            p=[0.60, 0.30, 0.05, 0.04, 0.01])
+        self.prob_loss_at_diag = self.hd_data.prob_loss_at_diag.sample()
         # FIXME: may be 2 or 3 if sw_art_disadv=1
         self.sw_incr_prob_loss_at_diag = 1
         self.higher_newp_less_engagement = rng.choice([True, False], p=[0.2, 0.8])
--- a/src/hivpy/hiv_diagnosis.py
+++ b/src/hivpy/hiv_diagnosis.py
@@ -1,0 +1,223 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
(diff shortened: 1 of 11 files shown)


src/hivpy/prep.py
```diff
@@ -0,0 +1,15 @@
+from enum import IntEnum
+
+
+class PrEPType(IntEnum):
+    Oral = 0
+    Injectable = 1
```
pineapple-cat: Two types of injectable to keep track of: CAB and LEN.

src/hivpy/hiv_status.py
```diff
@@ -416,3 +441,145 @@ def disease_and_diagnosis(disease_col, diagnosis_col, disease_rate, diagnosis_pr
...
+                                     sub_pop=pop.apply_bool_mask(diagnosed, general_pop))
+
+            # FIXME: should also include onart_tm1 and may need to be affected by date_most_recent_tb
+            # some people lost at diagnosis
+            lost = pop.transform_group([col.SEX_WORKER, pop.get_correct_column(col.NUM_PARTNERS, dt=1),
+                                        pop.get_correct_column(col.ADC, dt=1),
+                                        pop.get_correct_column(col.TB, dt=1),
+                                        pop.get_correct_column(col.NON_TB_WHO3, dt=1)],
+                                       self.calc_general_loss_at_diag,
+                                       sub_pop=pop.apply_bool_mask(diagnosed, general_pop))
```
pineapple-cat: Revisit order of operations and consider necessity of using values from the last time step.

src/hivpy/hiv_status.py
```diff
@@ -416,3 +441,145 @@ def disease_and_diagnosis(disease_col, diagnosis_col, disease_rate, diagnosis_pr
...
+        """
+        # tested population in primary infection
+        primary_pop = pop.get_sub_pop([(col.IN_PRIMARY_INFECTION, op.eq, True),
+                                       (col.LAST_TEST_DATE, op.eq, pop.date)])
+
+        if len(primary_pop) > 0:
+            # primary infection diagnosis outcomes
+            diagnosed = pop.transform_group([pop.get_correct_column(col.PREP_TYPE, dt=0),
+                                             pop.get_correct_column(col.PREP_TYPE, dt=1)],
+                                            self.calc_primary_diag_outcomes, sub_pop=primary_pop)
```
pineapple-cat: Requires discussion: this part seems more difficult to remove use of last time step from.

pineapple-cat: Add a variable to the PrEP module to keep track of when PrEP usage began.

src/hivpy/hiv_status.py
mmcleod89: Num partners condition should only be applied if `higher_newp_less_engagement` is true

mmcleod89: higher_newp_less_engagement is a sampled variable (20% chance true)

src/hivpy/hiv_status.py
mmcleod89: This is just to check that a person is taking injectable prep _and_ they started taking it before this timestep, so we would just need to check that the injectable prep start date is less than the current date.

src/hivpy/hiv_status.py
```diff
@@ -82,6 +91,20 @@ def __init__(self):
...
         self.other_adc_mortality_factor = rng.choice([1.5, 2, 3])
 
+        self.hiv_test_type = HIVTestType.Ab
+        self.test_sens_general = 0.98
+        self.test_sens_primary_ab = rng.choice([0.5, 0.75])
+        self.test_sens_prep_inj_primary_ab = rng.choice([0, 0.1])
+        # based on sens_tests_prep_inj in the SAS code
+        self.test_sens_prep_inj_primary_pcr = rng.choice([0.7, 0.5, 0.3, 0.2])
+        # FIXME: double-check if 0.05 being present twice is intentional
+        self.prob_loss_at_diag = rng.choice([0.01, 0.02, 0.04, 0.05, 0.05, 0.15, 0.30, 0.35, 0.50, 0.60])
```
pineapple-cat: This is intentional because (0.02, 0.05, 0.15, 0.35, 0.50) are the values and (0.60, 0.30, 0.05, 0.04, 0.01) are their respective sample probabilities.

src/hivpy/hiv_status.py
mmcleod89: If a person is diagnosed with HIV then they do not go onto injectable prep, but the injectable prep code comes before the diagnosis because it affects sensitivity. As a result, we don't change the sensitivity if the injectable prep variable has only been set _this_ timestep, so that it can be reversed later and be self consistent.

src/hivpy/hiv_status.py
```diff
@@ -426,3 +460,177 @@ def disease_and_diagnosis(disease_col, diagnosis_col, disease_rate, diagnosis_pr
...
+
+            # some people lost at diagnosis
+            lost = pop.transform_group([col.SEX_WORKER], self.calc_primary_loss_at_diag,
+                                       sub_pop=pop.apply_bool_mask(diagnosed, primary_pop))
+            pop.set_present_variable(col.UNDER_CARE, ~lost, sub_pop=pop.apply_bool_mask(diagnosed, primary_pop))
+
+        # remaining tested general population
+        general_pop = pop.get_sub_pop([(col.IN_PRIMARY_INFECTION, op.eq, False),
+                                       (col.LAST_TEST_DATE, op.eq, pop.date),
+                                       (col.HIV_DIAGNOSED, op.eq, False)])
```
mmcleod89: + HIV_STATUS = True

src/hivpy/hiv_status.py
```diff
@@ -14,6 +15,14 @@
...
 from . import output
 from .common import COND, SexType, opposite_sex, rng, timedelta
+from .prep import PrEPType
+
+
+# Ab (default), PCR (RNA VL), Ag/Ab
+class HIVTestType(IntEnum):
+    Ab = 0
+    PCR = 1
+    AgAb = 2
```
mmcleod89: Can we add comments for what these are 
@andrew-phillips-1 any useful information to provide for these acronyms?

mmcleod89:
Let's also move Diagnosis into its own module

src/hivpy/hiv_status.py
andrew-phillips-1: I suggest renaming PCR as NA for Nucleic Acid-based test

Ab - antibody only assay
NA - nucleic acid based 
AgAb - Antigen/antibody combined assay

We could change the order to 

Ab = 0
AgAb = 1
NA = 2

so that the higher the number the higher the sensitivity

pineapple-cat: Do we care about ordering by sensitivity more than ordering by how commonly you expect the test type to be used during simulations?

andrew-phillips-1: ok yes that's fine to keep as is then

src/hivpy/hiv_diagnosis.py
```diff
@@ -0,0 +1,230 @@
...
+        if len(general_pop) > 0:
+            # general diagnosis outcomes
+            diagnosed = pop.transform_group([col.PREP_TYPE, col.HIV_INFECTION_GE6M],
+                                            self.calc_general_diag_outcomes, sub_pop=general_pop)
+            # set outcomes
+            pop.set_present_variable(col.HIV_DIAGNOSED, diagnosed, general_pop)
+            pop.set_present_variable(col.HIV_DIAGNOSIS_DATE, pop.date,
+                                     sub_pop=pop.apply_bool_mask(diagnosed, general_pop))
+
+            # FIXME: should also include onart_tm1 and may need to be affected by date_most_recent_tb
```
mmcleod89, line 102: Is this in an issue anywhere?

pineapple-cat: Yes, it's included in #190.

mmcleod89 (commented):
Only two things:
- We should probably include more variation in test general loss at diagnosis for some variables that aren't checked like num stp 
- The variable for HIV infection > 6months isn't set anywhere. We could set it here using the hiv infection date if it's not going to be used anywhere else?

src/hivpy/hiv_diagnosis.py
mmcleod89: Super!

pineapple-cat:
All comments addressed; I just need to add a module overview and data variable section to the tutorial and this will be ready to merge.

mmcleod89 (approved):
This is fine; failing tests are not related to this PR and I will upload a fix today!

PR #191, opened 2024-06-18
Add tb_infection_date column and reset diagnosis variables

mmcleod89:
This should be usable for now; I do need to add better tests for TB and other diseases at some point, and double check how certain aspects of these disease variables are handled because it seems like they're calculated freshly every timestep in the SAS but I'm not sure how that works with infections that last a long time

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,7 +18,7 @@ packages = find:
 python_requires = >=3.7
 install_requires =
     pandas<1.4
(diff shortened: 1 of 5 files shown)


pineapple-cat (approved):
Changes look good and the unit test should be fine to add back in. Hopefully the CI will work now with the numpy version update in setup.

PR #195, opened 2024-07-25
PrEP eligibility selection

### LINKED ISSUE #85: PrEP Module Structure

### LINKED ISSUE #110: PrEP based HIV testing (without diagnosis)
Setting people who are interested in / eligible for PrEP for testing. 

SAS lines 4390-4527

PrEP variables can be dummied as this may pre-date any PrEP implementation.

pineapple-cat:
Addresses #85 and #110.


mmcleod89:
We should look into adding an `r_prep` column so that the random variable is only updated for people who are not yet eligible. (@andrew-phillips-1  we can discuss how this works on Wednesday)

mmcleod89:
`r_prep` only needs to be re-rolled if a person has gone from being prep eligible to ineligible, e.g. by losing a long term partner, long term partner going on ART, aging out, or changing sex worker status. 

In the SAS code prep eligibility is reset every time step so `r_prep` being retained helps to maintain the eligibility status for consistency. 

One approach to doing this without keeping eligibility for multiple time steps is to:
1. For everyone not currently marked prep eligible, randomly generate r_prep.
2. Set prep eligibility to false
3. Go through conditions to set prep eligibility (this should restore eligibility to people who shouldn't change)

src/hivpy/population.py
```diff
@@ -183,6 +183,12 @@ def get_sub_pop_intersection(self, subpop_1, subpop_2):
         """
         return pd.Index.intersection(subpop_1, subpop_2)
 
+    def get_sub_pop_union(self, subpop_1, subpop_2):
```
pineapple-cat: Make this variadic (so we're not calling unions in unions)? Or potentially just merge `risk_informed` and `suspect_risk` sub-pops into one sub-pop.

src/hivpy/prep.py
```diff
@@ -21,9 +26,218 @@ class PrEPModule:
 
     def __init__(self, **kwargs):
         # FIXME: move these to data file
```
pineapple-cat: I forgot to do this earlier, but I should probably address the data file FIXME as part of this PR.

src/hivpy/prep.py
```diff
@@ -20,10 +27,226 @@ class PrEPType(IntEnum):
...
+                # at_risk_ltp OR gen_ltp
+                prep_eligible_pop = pop.get_sub_pop_union(at_risk_ltp_pop, gen_ltp_pop)
+            # pregnant and lactating/breastfeeding women
+            elif self.prep_strategy == 16:
+                plw_pop = pop.get_sub_pop(AND(COND(col.HIV_DIAGNOSED, op.eq, False),
+                                              COND(col.SEX, op.eq, SexType.Female),
+                                              OR(COND(col.PREGNANT, op.eq, True),
+                                                 COND(col.BREASTFEEDING, op.eq, True))))
+                active_pop = pop.get_sub_pop(OR(COND(col.LONG_TERM_PARTNER, op.eq, True),
+                                                COND(col.LAST_STP_DATE, op.ge, pop.date - timedelta(months=9))))
```
mmcleod89: There's a discrepency here with core:
```
	if prep_any_strategy=16 then do;	* Pregnant and breastfeeding women (PLW) - new for MIHPSA Zimbabwe; *JAS Apr2023;
		* Note that there is a component of sexual behaviour in prep eligibility for pregnant and breastfeeding women;
      	if gender=2 and (pregnant=1 or breastfeeding=1) and 
		(newp ge 1 or (epdiag=1 and epart ne 1) or (ep=1 and epart ne 1 and (r < 0.05 or (r < 0.5 and epi=1)))) then prep_any_elig=1; 
		*( newp ge 1 or newp_tm1 ge 1 or newp_tm2 ge 1 or ep=1 ) then prep_any_elig=1; 	* replaced with code to match AGYW above Jan24;
	end;
```

The implementation which uses the previous two timesteps has been commented out; is there a reason why we are using this verion here?

mmcleod89 (commented):
Okay I'm basically happy with this; I made a few changes to tests to add in some people who were not meeting the criteria to check that the functions were rejecting ineligible people as well as acceptin eligible ones. 

I also changed some date related stuff where tm2 should translate to a date 6 months ago rather than 9 months ago, and tm1 should be three months ago. 

I think the only think I'm concerned about is the implementation of 16 being out of sync with what's currently in core. If this is how they want it implemented then I think we can go ahead now, otherwise it might be worth modifying in this PR

src/hivpy/prep.py
pineapple-cat: The current implementation is based on the eligibility code in `hiv_synthesis_laa_s.sas` on the `long_acting_treatment` branch:
```
	if prep_any_strategy=16 then do;	* Pregnant and breastfeeding women (PLW) - new for MIHPSA Zimbabwe; *JAS Apr2023;
		* Note that there is a component of sexual behaviour in prep eligibility for pregnant and breastfeeding women;
      	if gender=2 and (pregnant=1 or breastfeeding=1) and ( newp ge 1 or newp_tm1 ge 1 or newp_tm2 ge 1 or ep=1 ) then prep_any_elig=1; 
	end;
```
I've been using this branch because it contains some PrEP updates, such as the split of injectable PrEP into Cab and Len.

mmcleod89: Git blame puts these lines in both branches at about 5 months ago... @andrew-phillips-1 @jennifersmith203 do either of you know which version of this strategy we should be using in the python code? (As you two come up as the ones who have most recently edited this section!)

pineapple-cat:
We should double-check whether we need to use 9 or 6 months for most of the STP activity sections; the code uses information up to two time steps back, but all the comments specify "condomless sex in past 9 months", which implies to me the current time step includes activity from the past 3 months.

Could anyone clarify, @UCL/hiv-modelling?

```
    if prep_any_strategy=12 then do; 	* Any condomless sex in past 9 months;
    	if newp ge 1 or newp_tm1 ge 1 or newp_tm2 ge 1 or ep=1 then prep_any_elig=1; 
    end;

    if prep_any_strategy=13 then do; 	* Any condomless sex in past 9 months - women only; 
    	if gender=2 and ( newp ge 1 or newp_tm1 ge 1 or newp_tm2 ge 1 or ep=1 ) then prep_any_elig=1; 
    end;

	if prep_any_strategy=14 then do;	* as 4 but with newp_tm1 ge 1 also;	
    	r = rand('Uniform');
      	if (newp ge 1 or newp_tm1 ge 1 or (epdiag=1 and epart ne 1) or 
      	(gender=2 and 15 <= age < 50 and ep=1 and epart ne 1 and (r < 0.05 or (r < 0.5 and epi=1)))) then prep_any_elig=1; 
	end;
```

For example, I read this as strategies 12 and 13 going back 9 months, but strategy 14 only goes back 6 months. Is this the intention here?

mmcleod89:
> We should double-check whether we need to use 9 or 6 months for most of the STP activity sections; the code uses information up to two time steps back, but all the comments specify "condomless sex in past 9 months", which implies to me the current time step includes activity from the past 3 months.
> 
> Could anyone clarify, @UCL/hiv-modelling?
> 
> ```
>     if prep_any_strategy=12 then do; 	* Any condomless sex in past 9 months;
>     	if newp ge 1 or newp_tm1 ge 1 or newp_tm2 ge 1 or ep=1 then prep_any_elig=1; 
>     end;
> 
>     if prep_any_strategy=13 then do; 	* Any condomless sex in past 9 months - women only; 
>     	if gender=2 and ( newp ge 1 or newp_tm1 ge 1 or newp_tm2 ge 1 or ep=1 ) then prep_any_elig=1; 
>     end;
> 
> 	if prep_any_strategy=14 then do;	* as 4 but with newp_tm1 ge 1 also;	
>     	r = rand('Uniform');
>       	if (newp ge 1 or newp_tm1 ge 1 or (epdiag=1 and epart ne 1) or 
>       	(gender=2 and 15 <= age < 50 and ep=1 and epart ne 1 and (r < 0.05 or (r < 0.5 and epi=1)))) then prep_any_elig=1; 
> 	end;
> ```
> 
> For example, I read this as strategies 12 and 13 going back 9 months, but strategy 14 only goes back 6 months. Is this the intention here?

I think if this is the intent then we would still have to think carefully about how this gets implemented because the date set for the most recent short term partner is always set to the date of the timestep that the partner occurs in, which means that a partner from two timesteps ago will have a date of 6 months ago, and a partner from 3 timesteps ago (tm3) would have a date of 9 months ago. We could reconcile this by using _strictly_ less than 9 months ago, which would make the intent clear and still not inlcude people from three time-steps ago (as they would be considered 9 months and over).

mmcleod89:
Whilst what I wrote above is equivalent for a 3 month time step (<= 6months or < 9 months) this _isn't_ the case for other timesteps (e.g. 1 month) so it is still important to understand the _interpretation_ as well as the mechanics of it

andrew-phillips-1:
The intent is to look back to the time steps tm1 and tm2 so I suggest you ignore the comment if it is causing confusion.

On prep strategy 16, I suggest going with the version in Core (Jenny's version) rather than laa_s

mmcleod89:
@andrew-phillips-1 how should it generalise to a one month timestep though? Do you want to look at the last six months inclusive (the last 6 timesteps), or the last nine months non-inclusive (the last 8 timesteps). In a three month timestep both of theses scenarios look the same, but they generalise differently.

andrew-phillips-1:
Yes we have been thinking only in 3 month time step chunks.  I realize that when converting this to code that can be adapted for any time step then you have to make a decision as to how to define these things.  I would say that the two of you should just make a call on this which we go with.

andrew-phillips-1:
I think you can decide on how to convert.  The most important thing I guess is to try to be reasonably consistent with the approach to convert,

mmcleod89:
In that case I would propose going with the following solution: the date of the last short term partner is _strictly greater than_ 9 months. In this case, it be up to two timesteps ago in a 3 month timestep and up to 8 timesteps ago in a one month time step. This is consistent with the current SAS code as well as with the comment. @pineapple-cat I will make this change and push to the branch since I was the one fiddling with them to begin with! Can you update prep strategy 16? And then we're ready to go with the merge as far as I'm concerned :)

andrew-phillips-1:
Sounds good thanks Michael

mmcleod89 (commented):
I'm happy! :)

mmcleod89 (approved):
Clicked the wrong button so it became a comment instead of approval ...

Issue #196, opened 2024-07-31
Resettable State


mmcleod89:
We need to be able to reset the state to the year of intervention. Saving the dataframe should be easy enough, but we need to make sure that nothing in the _modules_ is going to cause a problem e.g. variables that change over time.

mmcleod89:
How easy is it to serialise custom classes in Python?

mmcleod89:
We need to be able to retain the state at the time of intervention and then continue with either option 0 (continuation of whatever the status quo is) or option X (some choice of intervention)

mmcleod89:
In `simulation.py`

- [ ] run simulation to year of intervention instead of end
- [ ]  make population deep copy and test that these are fully independent i.e. no references, changes in one (either to people or module variables) does not affect the other
- [ ] write a simple intervention function and call this from inside run once the copy has been made (and test!)
- [ ] run each simulation to stop date and capture output from both

PR #202, opened 2024-09-05
Year of intervention

dimitrasal:
The PR adds a feature where different policies can be implemented and the model predicts two plausible outcomes with and without intervention. In response to issue #196

### DIFF
--- a/hivpy.yaml
+++ b/hivpy.yaml
@@ -3,6 +3,7 @@ EXPERIMENT:
   start_year: 1989
   end_year: 2021
   time_interval_days: 90
+  intervention_year: 2000
   simulation_output_dir: output
   graph_outputs: []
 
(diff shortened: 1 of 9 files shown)


dimitrasal:
Need to discuss ways to:
 1) generalise the intervention function to allow for different measures, 
 2) make the code less repetitive

mmcleod89:
- No specific options code needs to be implemented as this is highly project specific
- Some simple intervention should exist for testing and demonstration purposes
- the intervention option should be able to be set in a yaml file, which would just be a number
- we should make a short tutorial explaining how to write interventions and alter aspects of the different modules

mmcleod89:
Could add:
- bool flag for repeating the intervention function every time step (false by default)
- a function for resetting / changing the option which then causes the inteverntion code to be called again. (They would then need to add a call to this in the loop, which we can explain in the documentation.)

mmcleod89 (approved):

PR #204, opened 2024-09-23
Starting PrEP usage

### LINKED ISSUE #205: Starting PrEP
Starting PrEP for the first time in two sub-populations:
- people who have explicitly tested to start PrEP
- people who are willing but have not explicitly tested to start PrEP

pineapple-cat:
Addresses #205.


src/hivpy/prep.py
```diff
@@ -52,13 +52,20 @@ def __init__(self, **kwargs):
         self.prep_willing_threshold = self.p_data.prep_willing_threshold
         self.prob_test_prep_start = self.p_data.prob_test_prep_start.sample()
         self.prob_prep_restart = self.p_data.prob_prep_restart.sample()
+        # FIXME: add to yaml
+        # probability of starting prep in people who are eligible and willing
+        # tested for HIV according to base rate of testing
+        self.prob_base_prep_start = rng.choice([0.05, 0.1, 0.2])
```
pineapple-cat: Leaving a reminder for myself to do this in this PR.

src/hivpy/prep.py
```diff
@@ -108,6 +125,38 @@ def get_suspect_risk_pop(self, pop: Population):
                                    COND(col.LTP_HIV_STATUS, op.eq, True),
                                    COND(col.R_PREP, op.lt, self.prob_suspect_risk_prep)))
 
+    def get_presumed_hiv_neg_pop(self, pop: Population):
```
pineapple-cat, line 135: #FIXME: This should have its own unit test.

src/hivpy/prep.py
```diff
@@ -356,3 +411,119 @@ def prep_eligibility(self, pop: Population):
...
+        for prep_type in sorted_dict:
+            willing = sorted_dict[prep_type]
+            if self.date >= self.date_prep_intro[prep_type] and willing:
+                starting_prep = prep_type
+                break
+
+        # outcomes
+        r = rng.uniform(size=size)
+        # FIXME: may need different probabilities for different prep types
+        starting = r < self.prob_base_prep_start
```
pineapple-cat: It's been confirmed that we want to retain the ability to apply different probabilities based on PrEP type, so this should be investigated in this PR as well.

src/hivpy/prep.py
```diff
@@ -108,6 +130,38 @@ def get_suspect_risk_pop(self, pop: Population):
                                    COND(col.LTP_HIV_STATUS, op.eq, True),
                                    COND(col.R_PREP, op.lt, self.prob_suspect_risk_prep)))
 
+    def get_presumed_hiv_neg_pop(self, pop: Population):
+        """
+        Return the sub-population that has been tested and is HIV positive but
+        received a false negative result.
+        """
+        false_neg_pop = pop.get_sub_pop(AND(COND(col.EVER_TESTED, op.eq, True),
+                                            COND(col.HIV_STATUS, op.eq, True)))
```
mmcleod89: Isn't this only a false negative if `HIV_DIAGNOSED` is equal to false as well?

pineapple-cat: You're right, that condition should be included here.

Edit: The reason it initially wasn't was because this sub-pop is only ever used in conjunction with the PrEP eligible sub-pop when starting PrEP, which already included the `HIV_DIAGNOSED == False` condition.

src/tests/test_prep.py
```diff
@@ -57,8 +60,8 @@ def test_prep_willingness():
     assert sum(pop.get_variable(col.PREP_CAB_WILLING, over_15s)) == 0
     assert sum(pop.get_variable(col.PREP_LEN_WILLING, over_15s)) == 0
     assert sum(pop.get_variable(col.PREP_VR_WILLING, over_15s)) == 0
-    # check for highest preference (ignore floats to discount nan values for under 15s)
-    assert all([i[0] is PrEPType.Oral for i in pop.data[col.PREP_PREF_RANKED] if type(i) is not float])
+    # check oral is highest preference (ignore unassigned 0 ranks for under 15s)
+    assert all(pop.data[col.PREP_ORAL_RANK] <= 1)
```
mmcleod89: Why not just check `==1` for over 15s?

pineapple-cat: This would be the more precise way of going about things, I just initially wanted to do this check in one assert when I first made the rank changes. I'll split this up to check for == 0 for under 15s and == 1 for over 15s.

mmcleod89:
@andrew-phillips-1 
I think we're having a difficult time interpreting something in the SAS code and therefore making it clear in the python code. 
The line in question is:
`if prep_any_ever ne 1 and tested=1 and (hiv=0 or (hiv=1 and unisensprep > eff_sens_vct)) then do;`
(currently 5166 in the SAS code on the main branch)

`unisensprep` is just a random number generated
`eff_sens_vct` appears to be a test sensitivity based on what test is available

What I'm confused about is that it appears to be modelling false negatives using a random variable, but don't false negatives already appear in the diagnosis part of the simulation? If we test all the HIV+ people who are selected for testing in a given time step, then the people who get diagnosed will be true positives, and the people who don't will be false negatives. Why is this code separate, is it modelling something different? Can this lead to people who are e.g. HIV+, get tested, get a (false) negative result, and still don't meet the criterion `unisensprep > eff_sens_vct`, and how would we interpret these people? 

Is this modelling a separate round of testing i.e. testing before starting prep separately (and possibly in addition to) other forms of testing? (Although it doesn't appear to lead to diagnosis?)

andrew-phillips-1:
Thanks @mmcleod89 @pineapple-cat I will try to have a think about this before our call tomorrow.  I've also let Jenny and Loveleen know.

andrew-phillips-1:
I think @mmcleod89 @pineapple-cat that this code is because the person has just tested in this period so we have not had a chance to get to the diagnosis part of the code.  So for this test we need to consider that it could be a false negative before PrEP can be started.  

(The fact that we are not using the same random draw when lower down we are seeing whether the person is diagnosed is not quite right though I think and it would be good to have that working in Python version.)

Am I misunderstanding ?

@LoveleenB-M @jennifersmith203

src/hivpy/prep.py
```diff
@@ -130,6 +130,8 @@ def get_suspect_risk_pop(self, pop: Population):
                                    COND(col.LTP_HIV_STATUS, op.eq, True),
                                    COND(col.R_PREP, op.lt, self.prob_suspect_risk_prep)))
 
+    # FIXME: this function may be removed if there are no issues with
+    # updating PrEP after HIV diagnosis during population evolution
```
pineapple-cat, line 134: Finding this out would include testing multiple modules together; where would it make the most sense to put this test?

mmcleod89: That's a good question; and it's tricky to say right now because we're not sure _what_ we would need to test yet to be able to figure this out. It might be possible to define some integration tests in a file of their own if we can say definitively what some behaviour of the simulation as a whole should look like when working together.

mmcleod89 (approved):
Looks good! We will need to look into updating CI to fix the test issue

PR #208, opened 2024-11-01
Continuous PrEP usage

### LINKED ISSUE #206: Re-starting, continuing, and switching PrEP
Track continuous PrEP usage and determine whether individuals on PrEP continue to take it, stop, or switch each time step. Individuals that are not currently on PrEP but have taken it before may also opt to re-start usage.

pineapple-cat:
Addresses #206.


src/hivpy/prep.py
```diff
@@ -94,6 +95,19 @@ def init_prep_variables(self, pop: Population):
...
 
+    # FIXME: should this function be in another module?
+    def get_vl_prevalence(self, pop: Population):
+        """
+        Return the prevalence of people between 15 and 50 years old with a viral load of over 1000.
+        Affects willingness to take PrEP.
+        """
+        gen_pop = len(pop.get_sub_pop([(col.AGE, op.ge, 15), (col.AGE, op.lt, 50)]))
+        # find prevalence of people with a viral load of over 1000
+        return (len(pop.get_sub_pop([(col.VIRAL_LOAD, op.gt, 1000),
```
pineapple-cat: We want to change this value to <= 3.0.

src/hivpy/prep.py
```diff
@@ -183,62 +214,49 @@ def set_prep_preference(self, pop: Population, date_intro, pref_beta, pref_col,
...
+
+    def prep_pref_ranks(self, pop: Population, sub_pop=None):
+        """
+        Rank PrEP preferences.
+        """
+        # get ranking outcomes
+        # FIXME: not sure if transform group is the best way to do this, but it works for now
+        pref_ranks = pop.transform_group([col.PREP_ORAL_PREF, col.PREP_CAB_PREF,
+                                          col.PREP_LEN_PREF, col.PREP_VR_PREF],
+                                         self.calc_prep_pref_ranks, sub_pop=sub_pop, use_size=False)
```
pineapple-cat: Probably worth using ``apply`` here instead.

mmcleod89 (approved):
Looks good to me for now; we will need to update though to have the prep code not necessarily run for everyone every time step as previously discussed

PR #218, opened 2024-11-26
Replace LTP_HIV_STATUS dummy column with LTP_STATUS column

mmcleod89:
LTP_HIV_STATUS and LTP_STATUS shouldn't both exist!


pineapple-cat (changes requested):
Just need to fix the remnants of `LTP_HIV_STATUS` in `test_prep.py` (`test_risk_informed_pop`, `test_suspect_risk_pop`, `test_prep_eligibility_continuity`, `test_prep_eligibility_women_only`, `test_prep_eligibility_all`) and I'm happy to sign this off.

pineapple-cat (approved):

mmcleod89:
Okay I looked into the issue and this is what I think was happening:
- The size of the array being calculated was correct: the total size should be 4 and we had two groups of size 2, both of which had its size calculated correctly. 
- However, I was concerned that we were performing a groupby on a prep_type when prep_type was coming up as `nan`. 
- PrEP_TYPE is by default set to `None` but _is not given a specific type_, somewhere it is undergoing a type conversion to `nan` which is a thing pandas can do which drives me insane
- I replaced the `None` in PrEP_TYPE code with a new entry in the enum called `NoPrEP` to make sure that pandas always has type information about the PrEP_TYPE. This does not appear to break any further code but you might want to take a look at it and check that you are happy and don't want to change anything else, especially given that the PrEP_TYPE is used for indexing.

mmcleod89:
I could also use your help in double checking the merge conflicts because I want to be sure that I'm not stepping on any toes in the prep module. 

Also the following lines in `prep.py` will need changing:

```
            # people who are either continuing or switching prep
            using_prep_pop = pop.apply_bool_mask(prep_types.notnull(), eligible)
            # people who are stopping prep
            stopping_prep_pop = pop.apply_bool_mask(prep_types.isnull(), eligible)
```
along with two others that feature `notnull()`, and I think some more in the prep tests. I think I can replace all of them them if we can handle the merge conflicts.

pineapple-cat (approved):
Thanks for sorting this out, looks like we got them all this time :)

PR #219, opened 2024-12-17
Create resistance mutations module

### LINKED ISSUE #211: Implement HIV progression on ART
EFFECT OF NACTIVE 

Major refactoring should simplify / compactify this code substantially.

pineapple-cat:
Addresses #211.


src/hivpy/resistance_mutations.py
```diff
@@ -0,0 +1,388 @@
...
+                                    pop.get_correct_column(col.ART_ADHERENCE, dt=0),
+                                    pop.get_correct_column(col.ART_ADHERENCE, dt=1),
+                                    col.MAX_VIRAL_LOAD],
+                                   self.calc_viral_load, sub_pop=sub_pop)
+
+        pop.set_present_variable(col.VIRAL_LOAD, viral_load, sub_pop)
+
+    def get_viral_load_matrix(self, max_viral_load):
+        # viral_load_matrix[active_drugs][cont_on_art][adherence]
+        # FIXME: is there a better way to do this?
```
pineapple-cat: This is ridiculously inefficient for population sizes of 100k+, but even switching to a static matrix containing a, b, and c values for a `a * max_viral_load + b + c * min_vl_on_art` calculation isn't much better because the main problem appears to be that a separate calculation is happening for every person in the population.

I think it looks like the vl calculation will have to be separated into more `transform_group`-able chunks to improve performance.

src/hivpy/population.py
```diff
@@ -118,12 +121,11 @@ def init_variable(self, name: str, init_val, n_prev_steps=0, data_type=None):
...
-        if (n_prev_steps == 0):
-            if data_type is not None:
-                self.data[name] = pd.Series([init_val]*self.size, dtype=data_type)
-            else:
-                self.data[name] = init_val
+        if data_type is not None:
+            self.data[name] = pd.Series([init_val]*self.size, dtype=data_type)
         else:
+            self.data[name] = init_val
+        if (n_prev_steps > 0):
```
mmcleod89: Let's put it back the way it was to avoid having the extra column created which could cause issues

src/hivpy/resistance_mutations.py
pineapple-cat: Using a static matrix and pre-calculating all matrix indices reduces the runtime of one iteration of viral load calculations on my machine by approximately 60% (46s > 18s with a sub-population size of 1 000 000).

src/hivpy/resistance_mutations.py
```diff
@@ -0,0 +1,422 @@
...
+        # recovery is hindered by a failing nnrti (or possibly insti) regimen
+        if (((person[col.ON_NEV] or person[col.ON_EFA]) or (self.failed_insti_hinders_cd4_recovery and person[col.ON_DOL]))
+                and not (person[col.ON_LPR] or person[col.ON_TAZ] or person[col.ON_DAR])
+                and person[col.NUM_ACTIVE_DRUGS] <= 2):
+            base_cd4_recovery_on_art = self.hindered_cd4_recovery
+        # recovery increases on pi
+        if person[col.ON_LPR] or person[col.ON_TAZ] or person[col.ON_DAR]:
+            base_cd4_recovery_on_art += self.cd4_recovery_pi_factor
+        # recovery decreases with age
+        base_cd4_recovery_on_art += (person[col.AGE] - 40) * - 0.3
```
mmcleod89:
```suggestion
        base_cd4_recovery_on_art -= (person[col.AGE] - 40) *  0.3
```
I just think `-= ... *` is easier to read that `+= ... * -`

mmcleod89 (approved):

Issue #212, opened 2024-11-19
Implement rise of mutations in people on ART due to low adherence etc.


mmcleod89:
Can be done in parallel with #211 if `newmut` is dummied 
Will also need to dummy on_drug 

May require starting the drugs/resistance dictionary

mmcleod89:
`NEW RESISTANCE MUTATIONS ARISING (and dominating)`

PR #221, opened 2025-02-17
Add rise of new resistance mutations (part i)

pineapple-cat:
Addresses #212.

Part i mutations: TAMs, RT, PR
Part ii mutations: IN, CA (+ cab / len tails)


src/hivpy/column_names.py
```diff
@@ -167,13 +167,12 @@
...
 RTTA_MUTATIONS = "rttams"                       # int: number of RT (reverse transcriptase) gene thymidine analog mutations (TAMs)
 RT184_MUTATION = "rt184m"                       # bool: True if an individual has the RT gene M184 mutation
-RT65_MUTATION = "rt65m"                         # bool: True if an individual has the RT gene K65 mutation
 RT151_MUTATION = "rt151m"                       # bool: True if an individual has the RT gene Q151 mutation
+RT65_MUTATION = "rt65m"                         # bool: True if an individual has the RT gene K65 mutation
 RT103_MUTATION = "rt103m"                       # bool: True if an individual has the RT gene K103 mutation
 RT181_MUTATION = "rt181m"                       # bool: True if an individual has the RT gene Y181 mutation
 RT190_MUTATION = "rt190m"                       # bool: True if an individual has the RT gene G190 mutation
 PR32_MUTATION = "pr32m"                         # bool: True if an individual has the PR (protease) gene P32 mutation
-PR33_MUTATION = "pr33m"                         # bool: True if an individual has the PR gene P33 mutation
```
pineapple-cat, line 176: I thought `PR33` was the only mutation missing from the `NEW RESISTANCE MUTATIONS ARISING (and dominating)` section, but upon further examination it seems `PR90` is also missing.

src/hivpy/resistance_mutations.py
```diff
@@ -422,38 +422,34 @@ def update_new_mutations_arising_art(self, pop: Population, sub_pop):
...
+
+            # m184
+            set_new_majority_mutation(col.RT184_MUTATION, [col.ON_3TC, col.ON_ISL, col.RT184_MUTATION],
+                                      self.calc_rt184m_outcomes)
+            # q151
+            set_new_majority_mutation(col.RT151_MUTATION, [col.ON_ZDV, col.RT151_MUTATION],
+                                      self.calc_rt151m_outcomes)
+            # k65
+            set_new_majority_mutation(col.RT65_MUTATION, [col.ON_TEN, col.ON_ZDV, col.RT65_MUTATION],
+                                      self.calc_rt65m_outcomes)
```
pineapple-cat, line 458: Is there a reason RT151 and RT65 check for their own majority presence when none of the other mutations do?

mmcleod89 (approved):
Happy with this, but did make a change to the additional test. This is so that it tests all possible combinations of resistance mutation presence (absence, minority, and majority) for the other two mutations and checks that there is no resistance developed it either of the others is present in majority, but that resistance does develop if they are absent _or in minority_.

src/hivpy/column_names.py
pineapple-cat: Removing `PR90` for now, but it may be added as an atazanavir mutation at some point.

src/hivpy/resistance_mutations.py
pineapple-cat: There doesn't seem to be a discernible function to this, so it will be removed for consistency.

PR #222, opened 2025-02-26
Add rise of new resistance mutations (part ii)

### LINKED ISSUE #212: Implement rise of mutations in people on ART due to low adherence etc.

pineapple-cat:
Addresses  #212.

Part i mutations: TAMs, RT, PR
Part ii mutations: IN, CA (+ cab / len tails)


src/hivpy/resistance_mutations.py
```diff
@@ -727,6 +752,37 @@ def calc_pr88m_outcomes(self, on_taz, size):
...
+        in_mutations |= r < prob_mutation
+
+        return in_mutations
+
+    def calc_ca66m_outcomes(self, on_ole, on_len, in_len_tail, in_len_post_tail, on_art, size):
+        """
+        Returns CA gene CA66 mutation outcomes.
+        """
+        prob_mutation = self.resist_rate_len if on_ole or on_len or in_len_tail \
+            or (in_len_post_tail and not on_art) else 0
```
mmcleod89: Probably better to put this in a sub-pop statement before the function rather than splitting up a transform group on all these variables

mmcleod89: Let's look into the performance impact a bit

tutorials/resistance_mutations.md
```diff
@@ -48,3 +48,41 @@ All mutations are tracked with a `MutationStatus` enum, except thymidine analog
...
+- *`taz_resist_rate`* - The rate of atazanavir resistance acquisition.
+- *`isl_resist_rate`* - The rate of islatravir resistance acquisition.
+- *`ten_resist_rate`* - The rate of tenofovir resistance acquisition.
+- *`dol_resist_rate`* - The rate of dolutegravir resistance acquisition.
+- *`len_resist_rate`* - The rate of lenacapavir resistance acquisition.
+- *`incr_len_resist`* - A multiplier for lenacapavir resistance acquisition.
+- *`cab_resist_factor`* - A multiplier for cabotegravir resistance acquisition.
+- *`viral_load_matrix`* - A matrix containing (`a`, `b`, `c`) float tuples used to calculate an individual's base viral load (`a * max_viral_load + b + c * min_vl_on_art`) for a given time step.
+- *`cd4_delta_matrix`* - A matrix containing floats used to calculate an individual's change in CD4 levels for a given time step.
+- *`new_mutation_matrix`* - A matrix containing floats used to calculate an individual's new mutation probability for a given time step.
```
pineapple-cat, line 88: @UCL/hiv-modelling As brought up today, I'd appreciate it if someone had a look at these variable descriptions to check that they make sense to you.

src/hivpy/resistance_mutations.py
pineapple-cat: So I created two slightly modified versions of `test_ca66m` and ran them with a population size of 100 000 (`update_new_mutations_arising_art` was called twice in each test). In one test everybody was infected with HIV, and in the other 5% of people were infected with HIV. I tested the original CA66 outcome calculation method, as well as a sub-pop method where I first selected all people either on oral Len, injectable Len, or in a Len tail, and then used that sub-pop with a modified CA66 calculation that featured fewer transform group columns.

Resulting average times from 5 runs of each test:
- `original_5%` = 3.5s
- `sub_pop_5%` = 3.4s
- `original_100%` = 22s
- `sub_pop_100%` = 23s

There doesn't seem to be much of a difference between the two methods.

tutorials/resistance_mutations.md
andrew-phillips-1: This seems reasonable, Emily.  I'm wondering if you would consider including in your comments in the code the names of the relevant parameter in the SAS code ?  I realize this would not be good for wider accessibility of the code but it would help us when we are transitioning.

pineapple-cat: Are there any specific modules or sections you think would particularly benefit from having the SAS names side-by-side? I think there may be some variables that don't have a direct 1:1 anymore, but most of them are still strongly based on the SAS code and typically have elements of the old names incorporated into the new names.

andrew-phillips-1: I don't think there are specific modules where it is most needed, and I think we will manage without this, but I thought it was worth us considering.

Any thoughts on this, @LoveleenB-M and @jennifersmith203 ?

src/hivpy/resistance_mutations.py
mmcleod89: Look good, we can keep things as they are then!

tutorials/resistance_mutations.md
mmcleod89: I think that creating a repository-wide dictionary between the python and the SAS might be quite a lot of work, especially for any variables which don't map directly, so it might not be the best use of time as we are heading towards the end of the project.

src/hivpy/prep.py
```diff
@@ -201,19 +201,20 @@ def get_presumed_hiv_neg_pop(self, pop: Population):
...
         takes some time to wear off (~3 months for cab and ~6 months for len).
         """
         return pop.get_sub_pop(OR(COND(col.PREP_TYPE, op.eq, PrEPType.Oral),
                                   COND(col.PREP_TYPE, op.eq, PrEPType.VaginalRing),
                                   AND(COND(col.PREP_TYPE, op.eq, PrEPType.Cabotegravir),
-                                      COND(col.LAST_PREP_USE_DATE, op.le, pop.date - timedelta(months=3))),
+                                      COND(col.LAST_PREP_USE_DATE, op.le, pop.date - self.cab_tail_length)),
                                   AND(COND(col.PREP_TYPE, op.eq, PrEPType.Lenacapavir),
-                                      COND(col.LAST_PREP_USE_DATE, op.le, pop.date - timedelta(months=6)))))
+                                      COND(col.LAST_PREP_USE_DATE, op.le, pop.date - self.len_tail_length))))
```
mmcleod89: I think that the time between prep injections is not necessarily equal to the length of the tail for a given drug injectable prep drug. (@andrew-phillips-1 ?) 
I think we will need separate variables for e.g. cab_inj_interval, len_inj_interval which are different from the tail length.

andrew-phillips-1: Yes probably better to have these as separate variables.

tutorials/resistance_mutations.md
andrew-phillips-1: OK yes - I agree.

mmcleod89 (approved):

